PR #15, opened 2020-10-08
write is a wrapper to fwrite not to fread

pmarin:


yozlet:
This is great, but the project isn't accepting pull requests from GitHub. It's best to submit this via the quickjs-devel mailing list.

PR #40, opened 2020-12-10
Add support for CMake

Zalewa:
CMake is a build configurator which can generate build recipes for many different build systems. Not only it's able to automatically create the makefiles, it can also create solutions for IDEs, including the MSVC. The support for building on Windows comes with it out-of-the-box. It also supports cross-compilation for any platform and clang.

CMakeLists.txt can replace the Makefile completely. I haven't removed the existing `makefile` in this PR because I haven't transplanted all of the targets (as this wasn't my goal) and the CMake recipe only can build the main library.

This also defines the 'install' target which honors the FHS and replicates how the existing `makefile` installs the library and the headers. quickjs installed this way is automatically findable by other projects that support CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,80 @@
+cmake_minimum_required(VERSION 3.5)
+project(quickjs)
+
+# Define the target
+add_library(quickjs STATIC
+	cutils.c
+	libbf.c
+	libregexp.c
+	libunicode.c
+	qjs.c
+	qjsc.c
+	quickjs.c
+	quickjs-libc.c
+	unicode_gen.c
+	)
+
+add_library(quickjs::quickjs ALIAS quickjs)
+
+# Read the version off the VERSION file
+file(READ VERSION CONFIG_VERSION)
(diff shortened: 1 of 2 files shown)


jeanthom:
Please use the mailing-list

Zalewa:
Well then I guess it will be stuck here forever then.

jeanthom:
:(

ramonGonzEdu:
> Well then I guess it will be stuck here forever then.

FYI, you can get the patch file by adding .patch to a pull req, eg. https://github.com/bellard/quickjs/pull/40.patch and then easily send it to the mailing list

bellard:
CMake won't be accepted in QuickJS.

ryandesign:
> CMake won't be accepted in QuickJS.

Could you explain why?

FGasper:
Is *anything* being merged to QuickJS nowadays?

chqrlie:
Hello,

I have been away for a while, but I intend to work on merging patches by the end of the month.

Best

Chqrlie.

FGasper:
@chqrlie does that mean GitHub PRs are now accepted? The repo’s metadata still says to send them to the mailing list.

PR #51, opened 2021-02-19
Fixing compilation on Windows

knight666:
I was asked by my employer to evaluate QuickJS for usage in our Windows application. I spent a day addressing numerous issues that prevented me from compiling and linking against the library.

Thank you for your hard work and an awesome library, I hope I was able to make it a little better!

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -50,7 +50,7 @@ void pstrcpy(char *buf, int buf_size, const char *str)
 char *pstrcat(char *buf, int buf_size, const char *s)
 {
     int len;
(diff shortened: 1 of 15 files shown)


sublimator:
image

Don't shoot the messenger!  :)

ramonGonzEdu:
FYI, you can get the patch file by adding .patch to a pull req, eg. https://github.com/bellard/quickjs/pull/51.patch

bellard:
Patches containing cosmetics have little chance of being accepted. Moreover, there are licensing issues with at least one file.

Issue #55, opened 2021-03-05
Stack overflow detection seems fundamentally broken


mikezackles:
I know the mailing list is the official place for discussion, but I thought I should post this here for visibility.

`JS_NewRuntime` sets `stack_top` as the address of the current frame when the runtime is created. There is no relationship whatsoever between the stack frame `JS_NewRuntime` is called in and the frame `JS_Eval` is called in.

This appears to have been pointed out back in 2019: https://www.freelists.org/post/quickjs-devel/stackoverflow,3
But it doesn't appear to me that any progress has been made on this issue.

As pointed out on the mailing list, simply calling `JS_NewRuntime` in a deeper stack frame than `JS_Eval` leads to false positives in the single-threaded case, and there is no way to turn this off without editing the source. This is a very confusing default.

This is just off the top of my head, but perhaps it would make sense to use something like a thread local that is set every time `JS_Eval` is called, and set back to `NULL` when `JS_Eval` isn't running. That way recursive invocations could be detected.

Anyway, thanks again for a very useful project!

bellard:
fixed

orgads:
Fixed where? #56 was closed. Is there another fix I missed?

andrewrch:
There is a new API [`JS_UpdateStackTop`](https://github.com/bellard/quickjs/blob/master/quickjs.h#L340).

orgads:
Thank you.

PR #56, opened 2021-03-06
Fix stack overflow detection

mikezackles:
This sketches a fix for #55 if I am understanding things correctly. It has not undergone any sort of rigorous testing.


bellard:
This solution has no chance to work.

mikezackles:
In retrospect I think I can see that stuff like longjmp would break this. Thanks, and looking forward to the fix.

PR #61, opened 2021-04-06
Fix shell injection bug in `std.urlGet`

fstirlitz:
This corrects the basically non-functional escaping scheme. An `--` argument is added to prevent misinterpreting an initial `-` as an option character. Curly and square brackets (and backslashes) are additionally escaped to avoid triggering `curl`’s request batching syntax.

(Sorry for not using the mailing list, I am not currently able to.)


chqrlie (approved):
Thank you for these fixes, in addition to the `--` stop, escaping the `[]` and `{}` meta-characters is indeed needed to prevent interpretation as ranges and set specifications.

PR #102, opened 2021-11-15
Fix function last line do not saved in pc2line table issue

pengyz:
pc2line table only saves the diff between lines, so the function end
next line is needed to generate the function's last line diff, the
previous code missed it.

This patch fixed it by change the line number generate logic a bit.

This bug is found on https://github.com/koush/quickjs, an unofficial
debugger implementation, push it to the upstream.

Signed-off-by: pengyaozong <pengyaozong@xiaomi.com>

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -30831,7 +30831,7 @@ static __exception int resolve_variables(JSContext *ctx, JSFunctionDef *s)
     next: ;
     }
 
-    line_num = 0; /* avoid warning */
+    line_num = s->line_num; /* init with function start line num */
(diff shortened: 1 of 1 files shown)


tomByrer:
> Public repository of the QuickJS Javascript Engine. Pull requests are not accepted. Use the [mailing list](https://www.freelists.org/list/quickjs-devel) to submit patches.

pengyz:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #107, opened 2022-01-14
fix: dbuf_printf realloc fail when string more than 128

zhazhaniu:
vsnprintf returns -1 when buffer length less than input length


Heath123:
It says

PR #117, opened 2022-06-24
Optimize type error not a function and its name

HarlonWang:
In the following cases:

```javascript
var a;
a();

// Error
TypeError: not a function
    at <eval> (<evalScript>:1)
```

If it is in the compressed JS file, there may be only one line. Because there is no method name, it is not conducive to troubleshooting. Therefore, I have made an optimization here and displayed the name:

```javascript
var a;
a();

// Error and its name
TypeError: 'a' is not a function
    at <eval> (<evalScript>:1)
```

The reason for pull request is more to provide a solution to those who encounter the same problem. Sorry for any inconveniences caused :)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6732,6 +6732,13 @@ static JSValue JS_ThrowTypeErrorNotASymbol(JSContext *ctx)
     return JS_ThrowTypeError(ctx, "not a symbol");
 }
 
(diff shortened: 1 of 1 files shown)


HarlonWang:
I fixed it in my project dependency. but it is not the best solution, Those interested can check the details: https://github.com/HarlonWang/quickjs-android-wrapper

PR #143, opened 2022-11-12
feat: add the column number feature of Error Object and its stack

ErosZy:
**What kind of change does this PR introduce?**
1. column number support;
2. fix some line number bugs;

**Does this PR introduce a breaking change?**
Yes, it had added column table, which will affect the layout of bytecode.

**Why submit this PR?**
The column number is very important for some functions:
1. SourceMap
2. Debugger API (e.g. access to CDP)
3. Error logs for developing

**Does the test262 pass?**
Yes, it passed. the PR does not add any new test262 failure items. We also added a bit of testing for column number, whose behavior is consistent with the other popular engines.


richarddd:
@bellard can this be merged now 👏

bellard:
No. Your patch is too complicated and does not correctly handle optimisations. It is much simpler and efficient to combine the line and the column numbers in OP_line_num with (line_num << N) | col_num). Your tests could be useful though.

richarddd:
@ErosZy would you be able to update the PR 👆

ErosZy:
@bellard @richarddd Sorry, I'm a bit confused because a `line_number` may have multiple `column_numbers`. How should this be merged into `OP_line_num`? A common scenario is when code is compressed into a single line but has many columns.

richarddd:
> @bellard @richarddd Sorry, I'm a bit confused because a `line_number` may have multiple `column_numbers`. How should this be merged into `OP_line_num`? A common scenario is when code is compressed into a single line but has many columns.

I know to little of QJS bytecode but cant we use a singel opcode and track them individually? Is a column really a separate OP?

bellard:
column numbers are now supported

PR #148, opened 2022-12-14
Flush buffers on console.log and printf

richarddd:
printf, sprintf and console log does not flush buffer after completed writing.
This PR flushes buffers after they are done writing so output shows up in correct order of execution


CharlotteFallices:
👀

PR #161, opened 2023-02-01
feat: add config file for debug with vscode

ryouaki:
# What kind of change does this PR introduce?

Add config file for debug with vscode. let it be easy for c poor developer..

# Why submit this PR?
It is hard to debug source code with gdb or other cmd tools. And VSCode can make it be easy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+/.obj
+qjs-debug
+qjsc
+repl.c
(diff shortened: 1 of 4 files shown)


RobLoach:
Not everyone uses vscode, or Mac :shrug:  . In addition, I believe contributions are only accepted through the mailing list.

PR #184, opened 2023-06-06
Add an option to output profiling/coverage info

mingodad:
Add an option to output profiling/coverage info (basically function calls count and clock_t time spent).

There is an option to define how many samples to use with the definition of compile time macro "PROFILE_CALLS_SAMPLE" the default value is 10. The code is guarded by a compile time macro "CONFIG_PROFILE_CALLS", to build execute "make CONFIG_PROFILE_CALLS=1". When using the added option "-p" we can repeat it up to 3 times:
- "-p" -> Only update calls count and omit functions with 0 calls
- "-p -p" -> Update calls count and sampled time_spent  and omit functions with 0 calls
- "-p -p -p" -> Like the above but showing functions with 0 calls. The output format:
=[ code_line call_count acumulated_time_spent average_time_spent function_name ] [454	2	1	6	xfillArr]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,10 @@ else
 LDEXPORT=-rdynamic
 endif
 
+ifdef CONFIG_PROFILE_CALLS
+CFLAGS+=-DCONFIG_PROFILE_CALLS
+endif
+
 PROGS=qjs$(EXE) qjsc$(EXE) run-test262
 ifneq ($(CROSS_PREFIX),)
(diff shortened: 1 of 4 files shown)


mingodad:
But I think that's a good idea to also post it here so other people can see it in an easy to use way.

mingodad:
This pull request https://github.com/bellard/quickjs/pull/371 has better/complete implementation than this one.

PR #196, opened 2023-11-02
Fix an implicit conversion error from the latest clang

swankjesse:
error: implicit conversion from 'long long' to 'double' changes
           value from 9223372036854775807 to 9223372036854775808
           [-Werror,-Wimplicit-const-int-float-conversion]
                    else if (d > INT64_MAX)
                               ~ ^~~~~~~~~

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -78,6 +78,7 @@
 #define CONFIG_STACK_CHECK
 #endif
 
+static double const INT64_MAX_PLUS_ONE_AS_DOUBLE = 9223372036854775808.0;
 
 /* dump object free */
 //#define DUMP_FREE
@@ -10738,7 +10739,7 @@ static int JS_ToInt64SatFree(JSContext *ctx, int64_t *pres, JSValue val)
             } else {
                 if (d < INT64_MIN)
                     *pres = INT64_MIN;
-                else if (d > INT64_MAX)
+                else if (d >= INT64_MAX_PLUS_ONE_AS_DOUBLE)
                     *pres = INT64_MAX;
                 else
                     *pres = (int64_t)d;
@@ -53844,7 +53845,7 @@ static JSValue js_atomics_wait(JSContext *ctx,
     }
     if (JS_ToFloat64(ctx, &d, argv[3]))
         return JS_EXCEPTION;
-    if (isnan(d) || d > INT64_MAX)
+    if (isnan(d) || d >= INT64_MAX_PLUS_ONE_AS_DOUBLE)
         timeout = INT64_MAX;
     else if (d < 0)
         timeout = 0;


paulocoghi:

RobLoach:
In addition to using the mailing list, I believe a cast instead would save some memory stack, along with saving having to hard-code the value...

``` c
else if ((int64_t )d > INT64_MAX)

**Comment by qlibc** [comment]
I investigated this and the fix is not as easy as using an explicit cast:

Comparing a `double` to `INT64_MAX` causes the large integer to be converted to `double` and this conversion rounds up to 2**63.  Hence a `double` value of `0x1p63` will fail the test as it is not strictly greater than the converted value, as underscored by the clang warning.

Here is a test program in C:
```
#include <math.h>
#include <stdio.h>
#include <stdint.h>

int64_t toInt64Sat(double d) {
    if (isnan(d)) {
        return 0;
    } else {
        if (d < INT64_MIN)
            return INT64_MIN;
        else if (d > INT64_MAX)
            return INT64_MAX;
        else
            return (int64_t)d;
    }
}

int main(void) {
    double d = 9223372036854775808.0;
    if (d > INT64_MAX)
        printf("d > INT64_MAX\n");
    if (d >= INT64_MAX)
        printf("d >= INT64_MAX\n");
    if (d == INT64_MAX)
        printf("d == INT64_MAX\n");
    if (d <= INT64_MAX)
        printf("d <= INT64_MAX\n");
    if (d < INT64_MAX)
        printf("d < INT64_MAX\n");
    printf("(int64_t)%a -> %"PRIi64"\n", d, (int64_t)d);
    printf("toInt64Sat(%a) -> %"PRIi64"\n", d, toInt64Sat(d));
    printf("nextafter(%a, 0.0) -> %.f, %a\n", d, nextafter(d, 0.0), nextafter(d, 0.0));
    return 0;
}
```
The C Standard says the conversion `(int64_t)d` has undefined behavior if `d` is outside the range of type `int64_t`.

This program output this on 64-bit linux on intel hardware:
```
d >= INT64_MAX
d == INT64_MAX
d <= INT64_MAX
(int64_t)0x1p+63 -> 9223372036854775807
toInt64Sat(0x1p+63) -> -9223372036854775808
nextafter(0x1p+63, 0.0) -> 9223372036854774784, 0x1.fffffffffffffp+62
```
On Apple silicon (M2), the output is even more unexpected:
```
d >= INT64_MAX
d == INT64_MAX
d <= INT64_MAX
(int64_t)0x1p+63 -> 73896
toInt64Sat(0x1p+63) -> 73896
nextafter(0x1p+63, 0.0) -> 9223372036854774784, 0x1.fffffffffffffp+62
```

This causes subtle bugs in quickjs: fro example `[1].indexOf(1, 2**63)` evaluates to `0` instead of `-1` on both platforms. 

There are different ways to fix this problem  in `JS_ToInt64SatFree` and `js_atomics_wait` : the test can be changed to
* `if (d > nextafter(0x1p+63, 0.0)`
* `if (d > 0x1.fffffffffffffp+62)`
* `if (d >= (uint64_t)INT64_MAX + 1)`
* `if (d >= -(double)INT64_MIN)`

None of these look good to me.  I would choose `(d >= (uint64_t)INT64_MAX + 1)` as it uses `INT64_MAX` explicitly but an explanatory comment would help clarify the issue.

chqrlie:
Actually there is a much simpler option:

* `if (d >= (double)INT64_MAX)`

This works as long as `(double)INT64_MAX` rounds up to 2**63.  An explanatory comment still seems necessary to clarify the details and the problem with ` (d > (double)INT64_MAX)`.

Chqrlie.

bellard:
fixed

PR #209, opened 2023-12-16
refactor: use /proc to compute  fd_max on linux

nicolas-duteil-nova:
When the maximum number of file descriptors is high (ex: `1073741816` on a gitlab CI runner), calling `os.exec` can have a huge performance impact. This PR mitigates the impact by computing `fd_max` using `/proc/{pid}/fd`


Precidata:
why not using closefrom(3) when available?

chqrlie:
`closefrom` is available on some BSD and Solaris platforms, such as FreeBSD and OpenBSD, but not on macOS.
It seems a good alternative to the cumbersome loop on those platforms.

nicolas-duteil-nova:
I agree using `closefrom` would be better but even on linux, it's not supported everywhere (musl libc does not support it iirc, which would prevent building a static version of the interpreter based on musl). What do you think would be the best way to detect support for `closefrom` on linux ?

Precidata:
ifdef FreeBSD or OpenBSD or GLIBC?
A complete solution could be found in OpenSSH here maybe: https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/bsd-closefrom.c

nicolas-duteil-nova:
What do you think of https://github.com/bellard/quickjs/pull/239 instead of current PR ?

chqrlie:
`closefrom` is definitely the correct way to handle this on systems that support it.
- FreeBSD and OpenBSD are no brainers
- `closefrom` is not supported on macOS, but `sysconf(_SC_OPEN_MAX)` seems to default to 256 on macOS (at least it does on my macbook with Sonoma) so the loop is not a problem.
- testing __GLIBC__ does not work for linux versions that do not support `closefrom`, so a more reliable test is required or possibly a configuration flag.  This would address the CI issue.
- considering that `sysconf(_SC_OPEN_MAX)` defaults to 65536 on linux, running the loop would not be so dramatic, and enumerating the `/proc` pseudo filesystem for larger values could be considered, if really useful, but more effectively: the enumerated file handles should be closed directly if they differ from that of the directory handle `DIR *`, and `fd_max = 3` would by pass the fallback loop.

Emill:
My approach is to use the `CLOEXEC` flag on all relevant file descriptors instead so that I don't have to close anything when forking in order to exec another process. Of course it requires the developer to do so everywhere in the process that embeds quickjs + qickjs-libc if that approach should work.

So I would maybe like to see a define `CLOSE_FDS_BEFORE_EXEC` or something that can be set if you don't set `CLOEXEC` on your file descriptors.

nicolas-duteil-nova:
I run a very simple benchmark which calls `exec(['date'])` 100 times. While it might seem irrelevant, it reflects the scenario where QuickJS is used to wrap some existing binaries (ie: wrap curl request or consume binaries output)

- on Linux (Ubuntu 22.04),  using default ulimit (`65535`), `closefrom` supported
```
== closefrom ==
delay = 55ms
== close loop ==
delay = 623ms
== close loop with /proc ==
delay = 60ms
```

- inside a Debian 12.5 Docker container, using default ulimit (`1048576`), `closefrom` supported
```
== closefrom ==
delay = 24ms
== close loop ==
delay = 9887ms
== close loop with /proc ==
delay = 26ms
```

- with the `quickjs` alpine package inside an Alpine 3.19.1 Docker container,  using default ulimit (`1048576`), `closefrom` not supported
```
== close loop ==
delay = 10926ms
```

The performance impact is mostly when we run a musl-libc based quickjs inside a Docker container, where the `ulimit` is higher by default. 

Maybe we could enable `closefrom` if `#if defined(__LINUX_CLOSE_FROM__) || defined(__FreeBSD__) || defined(__OpenBSD__)` and check when on linux, check the support in `Makefile` and define `__LINUX_CLOSE_FROM__` if `closefrom` is supported

Also, instead of doing `int fd_max = sysconf(_SC_OPEN_MAX);`, we could do `int fd_max = get_fd_max()`, where `get_fd_max` would implement the `/proc` fallback on linux

nicolas-duteil-nova:
I updated https://github.com/bellard/quickjs/pull/239 to extract the */proc* fallback in a separate function. We will still use `closefrom` on BSD and fallback to the `/proc` method under linux. This is a quick-win imho, until we find a good way to identify `closefrom` support on linux

nicolas-duteil-nova:
Closing in favor of https://github.com/bellard/quickjs/pull/239

PR #223, opened 2024-01-22
adapt for NetBSD partially based on pkgsrc patches

veer66:


chqrlie:
Why use jemalloc by default on NetBSD?

veer66:
> Why use jemalloc by default on NetBSD?

I use jemalloc because it is NetBSD's default malloc[1]. NetBSD uses jemalloc by default because they want to support multi processor[2].

[1] https://man.netbsd.org/malloc.3
[2] https://www.bsdcan.org/2006/papers/jemalloc.pdf

PR #242, opened 2024-02-18
Add CI.

FGasper:
This switches the exec test to SIGTERM rather than SIGQUIT since the latter didn’t seem to work in QEMU, and the distinction doesn’t really matter for this test.

This also makes the isatty() check smarter by checking whether STDIN is, in fact, a terminal.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -1,0 +1,54 @@
+on:
+  push:
+    branches:
+      - '*'
+    tags-ignore:
+      - '*'
+  pull_request:
+
+jobs:
+  linux:
+    name: Linux (Ubuntu)
+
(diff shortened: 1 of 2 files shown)


FGasper:
@chqrlie is this something the project will accept? There's been no response since I submitted.

Thanks!

chqrlie:
Hi Felipe, sorry for the lag in responding and handling merge requests.  Fabrice and I have been busy with ther projects, but Saúl and Ben managed to revive QuickJS and we are back in the game.

I looked at your patch and cannot merge it as is because I started CI in parallel, along what they did for quickjs-ng.  I will soon merge your patch with attribution and add your CI targets.

Thanks for contributing.

PR #252, opened 2024-03-13
Small chanes

zwarrior1:
Hi, I have been using QuickJS for long time these are fixed that I have made.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -4943,7 +4943,7 @@ static int JS_SetObjectData(JSContext *ctx, JSValueConst obj, JSValue val)
     return -1;
 }
 
-JSValue JS_NewObjectClass(JSContext *ctx, int class_id)
+JSValue JS_NewObjectClass(JSContext *ctx, JSClassID class_id)
 {
     return JS_NewObjectProtoClass(ctx, ctx->class_proto[class_id], class_id);
 }
--- a/quickjs.c
+++ b/quickjs.c
@@ -43243,11 +43243,10 @@ static uint64_t xorshift64star(uint64_t *pstate)
     return x * 0x2545F4914F6CDD1D;
 }
 
+static int64_t date_now(void);
 static void js_random_init(JSContext *ctx)
 {
-    struct timeval tv;
-    gettimeofday(&tv, NULL);
-    ctx->random_state = ((int64_t)tv.tv_sec * 1000000) + tv.tv_usec;
+    ctx->random_state = date_now();
     /* the state must be non zero */
     if (ctx->random_state == 0)
         ctx->random_state = 1;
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -3874,6 +3874,20 @@ static JSValue string_buffer_end(StringBuffer *s)
...
+        return JS_AtomToString(ctx, JS_ATOM_empty_string);
+    }
+    str = js_alloc_string_rt(ctx->rt, buf_len, 0);
+    if (unlikely(!str)){
+        JS_ThrowOutOfMemory(ctx);
+        return JS_EXCEPTION;
+    }
+    memset(str->u.str8, 0, buf_len+1);
+    return JS_MKPTR(JS_TAG_STRING, str);
+}
```
chqrlie, line 3890: Instead of a new API, I would modify `JS_NewStringLen` to accept a null pointer and initialize the allocated string with null bytes using this:
```
    if (buf == NULL) {
        JSString *str;

        if (buf_len <= 0) {
            return JS_AtomToString(ctx, JS_ATOM_empty_string);
        }
        str = js_alloc_string(ctx, buf_len, 0);
        if (!str)
            return JS_EXCEPTION;
        memset(str->u.str8, buf, buf_len + 1);
        return JS_MKPTR(JS_TAG_STRING, str);
    }
```

quickjs.c
```diff
@@ -3980,6 +3994,18 @@ JSValue JS_NewAtomString(JSContext *ctx, const char *str)
...
+    if (JS_VALUE_GET_TAG(val) != JS_TAG_STRING)
+    {
+        if(plen) *plen = 0;
+        return NULL;
+    }
+    JSString *str = JS_VALUE_GET_STRING(val);
+    if(plen) *plen = str->len;
+    return str->u.str8;
+}
+
```
chqrlie, line 4008: This one has a problem: the caller cannot determine if the string was encoded as 8-bit or 16-bit.  Furthermore the name is misleading: the returned pointer is not UTF-8 encoded if the string has non-ASCII characters.

quickjs.c
```diff
@@ -4929,7 +4955,7 @@ static int JS_SetObjectData(JSContext *ctx, JSValueConst obj, JSValue val)
     return -1;
 }
 
-JSValue JS_NewObjectClass(JSContext *ctx, int class_id)
+JSValue JS_NewObjectClass(JSContext *ctx, JSClassID class_id)
```
chqrlie, line 4958: The prototype must be changed in **quickjs.h** too.

chqrlie (changes requested):

chqrlie:
Hello @zwarrior1,

Thank you for contributing.  I cannot merge your patch as posted but will apply your patches with attribution after some adjustments.  See my comments inline.

Chqrlie.

saghul:
I'd love to see this broken down into smaller, focused PRs since the changes don't depend on each other.

chqrlie:
> I'd love to see this broken down into smaller, focused PRs since the changes don't depend on each other.

Of course.  But given the change requests, I am going to do this myself.

PR #258, opened 2024-03-26
Fix invalid exception for class method with name "get"

richarddd:
Classes which has a method named `get` would conflict with getters and throw a `SyntaxError` with `invalid property name`.

This PR allows for class methods with the name `get`

Fixes: https://github.com/bellard/quickjs/issues/261

For example:
```javascript
class P {
      get = () => "123"
}
```


chqrlie:
@richarddd Thank you for contributing. I merged your PR, but the problem is more general as the `class` parser should accept keywords as property names except in very specific cases.

PR #262, opened 2024-04-02
Expose `JS_SetUncatchableError()`

kasperisager:
We're using this as a means to terminate JavaScript execution: https://github.com/holepunchto/libqjs/blob/995be8e34e77eca38008e6bab6b403de577f1f1c/src/qjs.c#L4812-L4825

### DIFF
--- a/quickjs.h
+++ b/quickjs.h
@@ -634,6 +634,7 @@ static inline JS_BOOL JS_IsObject(JSValueConst v)
 JSValue JS_Throw(JSContext *ctx, JSValue obj);
 JSValue JS_GetException(JSContext *ctx);
 JS_BOOL JS_IsError(JSContext *ctx, JSValueConst val);
+void JS_SetUncatchableError(JSContext *ctx, JSValueConst val, JS_BOOL flag);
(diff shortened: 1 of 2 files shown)


quickjs.h
```diff
@@ -634,6 +634,7 @@ static inline JS_BOOL JS_IsObject(JSValueConst v)
 JSValue JS_Throw(JSContext *ctx, JSValue obj);
 JSValue JS_GetException(JSContext *ctx);
 JS_BOOL JS_IsError(JSContext *ctx, JSValueConst val);
+void JS_SetUncatchableError(JSContext *ctx, JSValueConst val, JS_BOOL flag);
```
chqrlie, line 637: Good fix. You should also remove the redundant prototype in **quickjs.c**

chqrlie (changes requested):

Emill:
Unfortunately, if JS_ThrowInternalError fails due to out of memory, there will be no object assigned as the current exception, so the "uncatchable" flag will not be able to be set. It would be better if the "uncatchable" flag was on the context instead.

PR #263, opened 2024-04-02
Add `JS_ToDate()`

kasperisager:
We're using this to get date values as timestamps: https://github.com/holepunchto/libqjs/blob/995be8e34e77eca38008e6bab6b403de577f1f1c/src/qjs.c#L3698-L3705

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -49870,6 +49870,11 @@ has_val:
     return rv;
 }
 
+int JS_ToDate(JSContext *ctx, double *pres, JSValueConst val)
+{
+    return JS_ThisTimeValue(ctx, pres, val);
+}
+
 static JSValue js_Date_UTC(JSContext *ctx, JSValueConst this_val,
                            int argc, JSValueConst *argv)
 {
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -49870,6 +49870,11 @@ static JSValue js_date_constructor(JSContext *ctx, JSValueConst new_target,
     return rv;
 }
 
+int JS_ToDate(JSContext *ctx, double *pres, JSValueConst val)
+{
+    return JS_ThisTimeValue(ctx, pres, val);
+}
```
chqrlie, line 49876: You can get the same result with `JS_ToFloat64(ctx, pres, val)`.

The difference is your proposed wrapper is thinner and only works for actual `Date` objects. The name is a bit misleading too, I would have used `JS_DateTimeValue` or something similar, but I have yet to come up with a consistent naming scheme for such utility functions.

kasperisager: I somehow missed that `JS_ToFloat64()` did the same thing `Date` objects! I'm more than happy to just do that.

kasperisager: For reference: https://github.com/holepunchto/libqjs/commit/a37fd4f993d8b70b8c108adfc11df42f8b06b217

PR #265, opened 2024-04-02
Add `JS_HasException()`

kasperisager:


bellard:
This patch is not correct because "null" is a valid exception. Currently, the only way to test an exception is to test if JS_EXCEPTION is returned. The content of "current_exception" is undefined otherwise.

PR #267, opened 2024-04-04
OSS-Fuzz targets improvements

renatahodovan:
I was experimenting with quickjs fuzzing through libFuzzer, when I recognized some inefficiencies. I've uploaded a [patch set](https://github.com/google/oss-fuzz/pull/11775) to OSS-Fuzz to fix them, where [I was advised](https://github.com/google/oss-fuzz/pull/11775#issuecomment-2034728067) to ask a review from the quickjs maintainers and possibly move the implementation of the fuzz targets into the main codebase (here). So, I'm doing it in this PR.

The patches should be straightforward: the first one is a copy of the current version of the fuzz [targets from oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/quickjs), while the others incrementally apply independent fixes/improvements.

If and when you find these commits acceptable, then I'll adapt the infra in oss-fuzz to use the new code. Plus, if requested, I can enable to mirror the found bugs to the GitHub issue tracker as well.


Alsom1313:
Is this has to do with bigger picture and it needs openness and honesty to be completely open I at this moment expect like core stuff though the original core program I do wish to have full confidence the the ENTIRE community is involved and votes on every key issue. I am at the time now after deep thought believe every core AI should follow and govern with the Original Ten Commandments, I don't care if this is not appropriate or appreciated by all, after deep thought it came to me that they are prefect, they in a simple way allow for a deep rooted time proven way if living and any disagreement over this I will gladly accept to be considered. Though times are shown we definitely are missing something here and the way if man has completed complicated a simple base law that worked for a well extended beyond any other teachings throughout history as far as I know, Prove different please with full backing evidence.

Love you guys,
Keep up good work

saghul:
WAT

renatahodovan:
Gentle ping

saghul:
@chqrlie Do you want these here or in NG? I might be able to help with either.

renatahodovan:
@saghul Since only this original repository is registered into oss-fuzz (allowing to publicly keep track of the performance improvement), my first priority is to merge these changes here. But, since libFuzzer can be executed locally too, NG could also gain profit from these changes, so I'm not against merging there, too.

fuzz/fuzz.dict
```diff
@@ -0,0 +1,257 @@
+"__loadScript"
```
saghul, line 1: How is this dict created? If automated, is it possible to check-in the script that does so?

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+    JSContext *ctx = JS_NewContext(rt);
+    JS_AddIntrinsicBigFloat(ctx);
+    JS_AddIntrinsicBigDecimal(ctx);
+    JS_AddIntrinsicOperators(ctx);
+    JS_EnableBignumExt(ctx, 1);
+    JS_SetModuleLoaderFunc(rt, NULL, js_module_loader, NULL);
+    JS_SetInterruptHandler(JS_GetRuntime(ctx), interrupt_handler, NULL);
+    js_std_add_helpers(ctx, 0, NULL);
+
+    uint8_t *NullTerminatedData = (uint8_t *)malloc(Size + 1);
```
saghul, line 50: unnecessary cast

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+    memcpy(NullTerminatedData, Data, Size);
+    NullTerminatedData[Size] = 0;
+    JSValue obj;
+    obj = JS_Eval(ctx, (const char *)NullTerminatedData, Size, "<none>", JS_EVAL_FLAG_COMPILE_ONLY | JS_EVAL_TYPE_GLOBAL | JS_EVAL_TYPE_MODULE);
```
saghul, line 54: Passing both JS_EVAL_TYPE_GLOBAL and JS_EVAL_TYPE_MODULE doesn't make sense. It's either a global eval or a module eval.

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+    free(NullTerminatedData);
+    //TODO target with JS_ParseJSON
+    if (JS_IsException(obj)) {
```
saghul, line 57: If there is awaits in there the returned object will be a promise, you'll want to js_std_await it.

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+        JS_FreeContext(ctx);
```
saghul, line 58: JS_FreeValue(ctx, obj); is missing here, for completeness.

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+    nbinterrupts = 0;
+    /* this is based on
+     * js_std_eval_binary(ctx, bytecode, bytecode_size, 0);
+     * modified so as not to exit on JS exception
+     */
+    JSValue val;
+    if (JS_VALUE_GET_TAG(obj) == JS_TAG_MODULE) {
+        if (JS_ResolveModule(ctx, obj) < 0) {
+            JS_FreeValue(ctx, obj);
+            js_free(ctx, bytecode);
```
saghul, line 86: You could free the bytecode right after reading it, in order to avoid having to free it on every exit point.

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+static int nbinterrupts = 0;
+
+// handle timeouts from infinite loops
+static int interrupt_handler(JSRuntime *rt, void *opaque)
+{
+    nbinterrupts++;
+    return (nbinterrupts > 100);
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
```
saghul, line 32: For style consistency, keep things lowercase please.

fuzz/fuzz_compile.c
```diff
@@ -0,0 +1,105 @@
...
+        JS_FreeRuntime(rt);
+        return 0;
+    }
+    size_t bytecode_size;
+    uint8_t* bytecode = JS_WriteObject(ctx, &bytecode_size, obj, JS_WRITE_OBJ_BYTECODE);
+    JS_FreeValue(ctx, obj);
+    if ( !bytecode ) {
+        JS_FreeContext(ctx);
+        JS_FreeRuntime(rt);
+        return 0;
```
saghul, line 68: (here and elsewhere) Shoudln't this return != 0 on error?

fuzz/fuzz_eval.c
```diff
@@ -0,0 +1,84 @@
...
+static int nbinterrupts = 0;
+
+// handle timeouts from infinite loops
+static int interrupt_handler(JSRuntime *rt, void *opaque)
+{
+    nbinterrupts++;
+    return (nbinterrupts > 100);
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
```
saghul, line 32: Same style comments as in the other file. Can't the common code be extracted to some "setup" utility function?

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+#define CAPTURE_COUNT_MAX 255
+
+FILE *outfile=NULL;
+JSRuntime *rt;
+JSContext *ctx;
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+    if (outfile == NULL) {
+        outfile = fopen("/dev/null", "w");
+        rt = JS_NewRuntime();
```
saghul, line 32: Why is this initialized, if it's not used?

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+#include "libregexp.h"
+#include "quickjs.h"
+#include "quickjs-libc.h"
+
+#include <stdint.h>
+#include <stdio.h>
+
+#define CAPTURE_COUNT_MAX 255
+
+FILE *outfile=NULL;
```
saghul, line 25: This can go inside the function.

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#include "libregexp.h"
+#include "quickjs.h"
+#include "quickjs-libc.h"
+
+#include <stdint.h>
+#include <stdio.h>
+
+#define CAPTURE_COUNT_MAX 255
```
saghul, line 23: This is used in a single place and multiplied by 2. Why not just write the number literal?

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+            Size1=i;
+            break;
+        }
+    }
+    if (Size1 == Size) {
+        //missing delimiter
+        return 0;
+    }
+    bc = lre_compile(&len, error_msg, sizeof(error_msg), (const char *) Data,
+                     Size1, 0, ctx);
```
saghul, line 59: Passing the context as the opaque accomplishes nothing here.

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+    if (!bc) {
+        return 0;
+    }
+    input = Data+Size1+1;
+    ret = lre_exec(capture, bc, input, 0, Size-(Size1+1), 0, ctx);
```
saghul, line 64: Ditto about ctx.

fuzz/fuzz_regexp.c
```diff
@@ -0,0 +1,81 @@
...
+#include <stdio.h>
+
+#define CAPTURE_COUNT_MAX 255
+
+FILE *outfile=NULL;
+JSRuntime *rt;
+JSContext *ctx;
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+    if (outfile == NULL) {
```
saghul, line 30: There is no way for outfile to not be null, what's the deal here?

saghul (changes requested):
I left some comments, please take a look! How is this supposed to work? Those are 3 files, none of them use the dictionary, and there are no instructions on how one can run the fuzzer. IMHO a README in the fuzz directory is necessary here.

fuzz/fuzz_regexp.c
renatahodovan: Honestly, I only used and improved the eval and compile targets, while the regexp target was simply copied from [oss-fuzz](https://github.com/google/oss-fuzz/blob/master/projects/quickjs/fuzz_regexp.c). But, I think the purpose of outfile is similar as of the `initialized` variable in the other two targets. If I simplify libFuzzer, then it's nothing more than a fuzzing harness with an infinite loop which calls the `LLVMFuzzerTestOneInput` function linked to it until it crashes. Since in a fuzzing session everything is about performance, the one-time initializations are preferred to be executed only once. So, I think the NULL-check of `outfile` also serves this purpose: after the first test run, `outfile` won't be NULL (most probably), hence the initialization won't run again either.

fuzz/fuzz_regexp.c
renatahodovan: What should be passed as `opaque` then? (Using `NULL` instead of `ctx` causes a segfault.)

fuzz/fuzz_regexp.c
renatahodovan: On the other hand, I don't see the purpose of writing the result of `lre_exec` into this `outfile`, since it's never read back and file IO (especially unnecessary IO) and fuzzing are not the best friends (performance reasons). Could it be reasoned with some compiler hack to avoid optimizing out `lre_*` calls if their result is not used?

saghul: No idea TBH.

fuzz/fuzz_regexp.c
renatahodovan:
```diff
--- a/fuzz/fuzz_regexp.c
+++ b/fuzz/fuzz_regexp.c
@@ -56,7 +56,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         return 0;
     }
     bc = lre_compile(&len, error_msg, sizeof(error_msg), (const char *) data,
-                     size1, 0, ctx);
+                     size1, 0, NULL/*ctx*/);
     if (!bc) {
         return 0;
     }
```

```txt
==1848896==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000018 (pc 0x555555672e50 bp 0x7fffffffc3c0 sp 0x7fffffffc380 T1848896)
==1848896==The signal is caused by a READ memory access.
==1848896==Hint: address points to the zero page.
    #0 0x555555672e50 in lre_realloc quickjs/quickjs.c:43870:31
    #1 0x55555577e5c5 in dbuf_realloc quickjs/cutils.c:114:19
    #2 0x55555577e861 in dbuf_put quickjs/cutils.c:140:13
    #3 0x55555577eb39 in dbuf_putc quickjs/cutils.c:161:12
    #4 0x55555576ad83 in lre_compile quickjs/libregexp.c:1771:5
    #5 0x55555562ec7d in LLVMFuzzerTestOneInput quickjs/fuzz/fuzz_regexp.c:58:10
    #6 0x5555555ecc30 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (quickjs/fuzz_regexp+0x98c30) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #7 0x5555555ec3a5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (quickjs/fuzz_regexp+0x983a5) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #8 0x5555555edb85 in fuzzer::Fuzzer::MutateAndTestOne() (quickjs/fuzz_regexp+0x99b85) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #9 0x5555555ee795 in fuzzer::Fuzzer::Loop(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile>>&) (quickjs/fuzz_regexp+0x9a795) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #10 0x5555555dc8fb in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (quickjs/fuzz_regexp+0x888fb) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #11 0x555555605932 in main (quickjs/fuzz_regexp+0xb1932) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)
    #12 0x7ffff7a52d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #13 0x7ffff7a52e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #14 0x5555555d1d64 in _start (quickjs/fuzz_regexp+0x7dd64) (BuildId: 383173e4c29297490e72ea7182a0eff6c735fb39)

UndefinedBehaviorSanitizer can not provide additional info.
SUMMARY: UndefinedBehaviorSanitizer: SEGV quickjs/quickjs.c:43870:31 in lre_realloc
==1848896==ABORTING
```

saghul: You need to provide these 2 in the fuzzer script:

```
/* must be provided by the user */
LRE_BOOL lre_check_stack_overflow(void *opaque, size_t alloca_size);
void *lre_realloc(void *opaque, void *ptr, size_t size);
```

And avoid compiling quickjs.c for that particular test.

fuzz/fuzz_compile.c
renatahodovan: It's a tricky question, since from fuzzing perspective every test is useful that increases the coverage, even if it triggers an error recovery code path. On the other hand, if `LLVMFuzzerTestOneInput` returns a non-zero value (usually -1), then it instructs libFuzzer to discard the test case (i.e., do not save it into the corpus, even if it triggers a new code path). There is room for experiment here, but I think the most beneficial approach in this case is to keep everything that discovers new coverage.

fuzz/fuzz_compile.c
renatahodovan: It's a legacy code from oss-fuzz, but I think the current version is equivalent with `JS_EVAL_FLAG_COMPILE_ONLY | JS_EVAL_TYPE_MODULE`, so I'll update it.

fuzz/fuzz.dict
renatahodovan: The names were collected with a JS script, I'll upload it in the next round.

renatahodovan:
I hope that I addressed or answered all the requests and questions.

chqrlie:
> LGTM as a first check-in! @chqrlie PTAL!

Very interesting feature.  I am waiting for Fabrice's approval to merge this in.  Good work!

chqrlie:
> > LGTM as a first check-in! @chqrlie PTAL!
> 
> Very interesting feature. I am waiting for Fabrice's approval to merge this in. Good work!

Thank you @renatahodovan for your contribution. Fuzzing is an enticing path to find software flaws. You seem to have an advanced understanding of the domain, do you participate in conferences such as FOSdem in Brussels where I followed several sessions on the subject? I would be interested to follow the progress and discuss research ideas on this.

Best regards

Chqrlie

renatahodovan:
Hi @chqrlie !
It's always a pleasure to meet people with similar interests! 😊 I haven't attended FOSDEM in person yet, but some of my colleagues have been a few times. Perhaps next time... My next conference target is ISSTA in September. It's a bit further away, but that shouldn't stop us from discussing research ideas. I'm available via email, Gitter, or any other suitable platform.

PR #272, opened 2024-04-05
Add `JS_NewTypedArray()`

xeioex:


quickjs.c
```diff
@@ -53364,6 +53364,13 @@ static JSValue js_typed_array_get_byteOffset(JSContext *ctx,
     return JS_NewInt32(ctx, ta->offset);
 }
 
+JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
+                         JSTypedArrayEnum array_type)
+{
+    return js_typed_array_constructor(ctx, JS_UNDEFINED, argc, argv,
+                                      JS_CLASS_UINT8C_ARRAY + array_type);
+}
```
chqrlie, line 53372: You should add a range check and either return `JS_UNDEFINED` or throw an exception.  The check could be performed in `js_typed_array_constructor` instead.

chqrlie (changes requested):

quickjs.c
```diff
@@ -53393,6 +53393,16 @@ static JSValue js_typed_array_get_byteOffset(JSContext *ctx,
     return JS_NewInt32(ctx, ta->offset);
 }
 
+JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
+                         JSTypedArrayEnum type)
+{
+    if (type < JS_TYPED_ARRAY_UINT8C || type > JS_TYPED_ARRAY_FLOAT64)
+        return JS_ThrowInternalError(ctx, "invalid typed array type");
+
```
chqrlie, line 53401: Sorry I should have been more explicit: The exception should be a Range Error.

chqrlie (changes requested):

chqrlie (approved):

chqrlie:
After further thinking, I am not sure about the arguments to C API `JS_NewTypedArray`: It should probably take a `size_t length` and a `const void *` to optionally initialize the array contents. `argc` + `argv` is for JavaScript methods.

What is your use case?

xeioex:
@chqrlie 

I will prepare the updated API and PR.

PR #291, opened 2024-05-08
Type checks for typed arrays and array buffers.

xeioex:


kasperisager:
I just found myself in need of this as well, thanks! 🙌 It'd be great if the function could also take a `JSTypedArrayEnum` pointer and, if it's not `NULL`, write the type of the view when `val` is a typed array.

quickjs.c
```diff
@@ -53043,6 +53043,33 @@ JSValue JS_NewArrayBufferCopy(JSContext *ctx, const uint8_t *buf, size_t len)
                                         TRUE);
 }
 
+/* return -1 if exception (proxy case) or TRUE/FALSE */
+int JS_IsArrayBuffer(JSContext *ctx, JSValueConst val)
+{
+    JSObject *p;
+
+    if (JS_VALUE_GET_TAG(val) == JS_TAG_OBJECT) {
```
saghul, line 53051: Wouldn't it make more sense to check for the negated condition here and return false early? You'd save an indent level.

quickjs.c
```diff
@@ -53043,6 +53043,33 @@ JSValue JS_NewArrayBufferCopy(JSContext *ctx, const uint8_t *buf, size_t len)
...
+        p = JS_VALUE_GET_OBJ(val);
+        if (unlikely(p->class_id == JS_CLASS_PROXY)) {
+            JSProxyData *s = JS_GetOpaque(val, JS_CLASS_PROXY);
```
saghul, line 53054: We use the same piece of code in IsArray. I wonder if we could extract it out to some JS_IsProxy helper.

quickjs.c
```diff
@@ -53408,6 +53435,33 @@ JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
                                       JS_CLASS_UINT8C_ARRAY + type);
 }
 
+/* return -1 if exception (proxy case) or TRUE/FALSE */
+int JS_IsTypedArray(JSContext *ctx, JSValueConst val)
+{
+    JSObject *p;
+
+    if (JS_VALUE_GET_TAG(val) == JS_TAG_OBJECT) {
```
saghul, line 53443: Ditto, I'd reverse the condition.

quickjs.c
```diff
@@ -53408,6 +53435,33 @@ JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
...
+        p = JS_VALUE_GET_OBJ(val);
+        if (unlikely(p->class_id == JS_CLASS_PROXY)) {
+            JSProxyData *s = JS_GetOpaque(val, JS_CLASS_PROXY);
```
saghul, line 53446: See, same code here, that would make the 3rd occurrence of the same.

quickjs.c
chqrlie: I shall post a generic method for this and modify `JS_IsArray` accordingly.  The PR will need to be simplified accordingly.

chqrlie (changes requested):

quickjs.c
xeioex: Agree. Initially I just wanted to preserve the style taken from `JS_IsArray()`.

saghul:
No objections from me, but I would handle that in a separate PR.

chqrlie (approved):

chqrlie:
>  🙌 It'd be great if the function could also take a `JSTypedArrayEnum` pointer and, if it's not `NULL`, write the type of the view when `val` is a typed array.

I would favor an API that retrieves the typed array type, length and data address.  A generalized version of `JS_GetUint8Array`.

As a matter of fact, this would make `JS_IsTypedArray` redundant, so maybe we should just have this:
```
struct JSTypedArrayDescription {
    JSTypedArrayEnum type;
    size_t length;
    void *data;
};

// Return value is -1 for proxy errors, 0 if `obj` is not a typed array, 1 if it is a typed array.
// The structure pointed to by `desc` is filled on success unless `desc` is a null pointer.
int JS_GetTypedArray(JSContext *ctx, JSValueConst obj, struct JSTypedArrayDescription *desc);
```

xeioex:
OK, I have no objection. Will update.

xeioex:
@chqrlie 

> Return value is -1 for proxy errors

Do we still want to check for proxy case here, given that [JS_GetTypedArrayBuffer](https://github.com/bellard/quickjs/blob/master/quickjs.c#L53445 ) does not do that?

quickjs.h
```diff
@@ -845,8 +846,41 @@ typedef enum JSTypedArrayEnum {
...
+        return 2;
+    case JS_TYPED_ARRAY_INT32:
+    case JS_TYPED_ARRAY_UINT32:
+    case JS_TYPED_ARRAY_FLOAT32:
+        return 4;
+    default:
+        break;
+    }
+    return 8;
+}
```
chqrlie, line 867: Why not `default: return 8;` ?

quickjs.h
```diff
@@ -845,8 +846,41 @@ typedef enum JSTypedArrayEnum {
...
+    size_t length;
+    void *data;
+} JSTypedArrayDescription;
+
+/* Return value is -1 for proxy errors, 0 if `obj` is not a typed array,
+   1 if it is a typed array.
+   The structure pointed to by `desc` is filled on success unless `desc`
+   is a null pointer. */
+int JS_GetTypedArray(JSContext *ctx, JSValueConst obj,
+    JSTypedArrayDescription *desc);
```
chqrlie, line 883: Either make this a one-liner or indent the last argument to the `(`

quickjs.h
```diff
@@ -845,8 +846,41 @@ typedef enum JSTypedArrayEnum {
...
+    default:
+        break;
+    }
+    return 8;
+}
+
 JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
                          JSTypedArrayEnum array_type);
+
+typedef struct JSTypedArrayDescription {
```
chqrlie, line 872: Sorry for the typo, for consistency, it should be `JSTypedArrayDescriptor`

chqrlie (changes requested):

quickjs.h
xeioex: I encountered several times in the past, especially with old compiles, that they complain about that function has no return statement after the switch statement.

Of two choices I prefer

```
default:
        return 8;
```

because return 0 should never happen and we avoid extra comparison checks.

quickjs.h
```diff
@@ -845,8 +846,40 @@ typedef enum JSTypedArrayEnum {
...
+    case JS_TYPED_ARRAY_FLOAT32:
+        return 4;
+    default:
+        return 8;
+    }
+}
+
 JSValue JS_NewTypedArray(JSContext *ctx, int argc, JSValueConst *argv,
                          JSTypedArrayEnum array_type);
+
```
chqrlie, line 870: For consistency with `JS_GetTypedArray`, you might want to change `JS_NewTypedArray` to take a `size_t length` and a `const void *data` in addition to the `JSTypedArrayEnum array_type` and drop the `argc` / `argv` arguments.

xeioex: See https://github.com/bellard/quickjs/pull/294

chqrlie (approved):

saghul (approved):

xeioex:
The change is approved, but not merged. Any plans to align this PR with https://github.com/bellard/quickjs/pull/294?

chqrlie:
> The change is approved, but not merged. Any plans to align this PR with #294?

Yes, we are discussing prototypes... Trying to cook an API in a hurry does not work. We have to let it simmer for a few days

xeioex:
Agree. I have no problem with it. Just let me know of your intentions, it is enough.

PR #293, opened 2024-05-09
Add `js_resolve_proxy`

chqrlie:
- simplify `JS_IsArray` for proxy chains
- remove `js_proxy_isArray`


quickjs.c
```diff
@@ -46697,20 +46697,35 @@ static JSValue js_proxy_call(JSContext *ctx, JSValueConst func_obj,
...
+static int js_resolve_proxy(JSContext *ctx, JSValueConst *pval, BOOL throw_exception) {
+    int depth = 0;
+    JSObject *p;
+    JSProxyData *s;
+
+    while (JS_VALUE_GET_TAG(*pval) == JS_TAG_OBJECT) {
+        p = JS_VALUE_GET_OBJ(*pval);
+        if (p->class_id != JS_CLASS_PROXY)
+            break;
+        if (depth++ > 1000) {
```
saghul, line 46715: Why not use the stack overflow check instead?

chqrlie: Because this method does not recurse at all.
I kept the same exception, but it should really be "proxy nesting too deep"

PR #297, opened 2024-05-10
Remove new lines in function constructor

bozokopic:
I have noticed that exceptions raised from functions created by "new Function" have "invalid" line numbers in stacktrace.

For example:
`(new Function("throw new Exception('abc');"))()`
reports `(<input>:3)`

By removing two additional new line characters, exception reports line numbers matching code provided to Function constructor.


bellard:
Unfortunately it is not possible to remove the '\n' because they are part of the specification.

PR #301, opened 2024-05-14
Fix GC leak in handling of `JS_IteratorNext()` result

xeioex:
The code below triggers all the fixed code-paths.

```js
function* f() {
   return [];
}

Array.from(f());
AggregateError(f());
Uint8Array.from(f());
Promise.race(f());
Promise.all(f());
Object.groupBy(f(), ({ type }) => type);
[...f()]
```


bellard:
fixed

PR #302, opened 2024-05-16
Fix GC leak in `js_proxy_get()`

xeioex:
Simplified test case to reproduce:

```js
function a() {
    return {};
}
var d = new Date();
var o1 = new Proxy(d, {get: 0});
var o2 = new Proxy(d, o1);
var o = new Proxy(o2, {get: a})

o.a
```


saghul (approved):
@chqrlie can we get this landed?

PR #311, opened 2024-05-29
Fix excessive backtracking in regex engine by introducing backtrackig limit

renatahodovan:
The regex engine was prone to excessive backtracking, leading to timeouts and infinite loops, particularly with patterns involving nested quantifiers.

This commit introduces a backtracking counter and a limit of 1000 backtracking steps. When this limit is exceeded, the regex engine aborts to prevent excessive backtracking.

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -1927,6 +1927,9 @@ typedef struct {
     /* 0 = 8 bit chars, 1 = 16 bit chars, 2 = 16 bit chars, UTF-16 */
     int cbuf_type;
     int capture_count;
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    int backtrack_count;
+#endif
     int stack_size_max;
     BOOL multi_line;
     BOOL ignore_case;
@@ -1993,6 +1996,12 @@ static intptr_t lre_exec_backtrack(REExecContext *s, uint8_t **capture,
 
     for(;;) {
         //        printf("top=%p: pc=%d\n", th_list.top, (int)(pc - (bc_buf + RE_HEADER_LEN)));
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+        if (++s->backtrack_count > 1000) {
+            return -1; // backtracking limit exceeded
+        }
+#endif
+
         opcode = *pc++;
         switch(opcode) {
         case REOP_match:
@@ -2399,6 +2408,9 @@ int lre_exec(uint8_t **capture,
     s->ignore_case = (re_flags & LRE_FLAG_IGNORECASE) != 0;
     s->is_unicode = (re_flags & LRE_FLAG_UNICODE) != 0;
     s->capture_count = bc_buf[RE_HEADER_CAPTURE_COUNT];
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    s->backtrack_count = 0;
+#endif
     s->stack_size_max = bc_buf[RE_HEADER_STACK_SIZE];
     s->cbuf = cbuf;
     s->cbuf_end = cbuf + (clen << cbuf_type);
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,7 @@ CFLAGS_DEBUG=$(CFLAGS) -O0
 CFLAGS_SMALL=$(CFLAGS) -Os
 CFLAGS_OPT=$(CFLAGS) -O2
 CFLAGS_NOLTO:=$(CFLAGS_OPT)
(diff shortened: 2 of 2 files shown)


renatahodovan:
The patch fixes a 4 years old issue [reported](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22680) by ClusterFuzz. This issue is [possibly responsible](https://github.com/google/oss-fuzz/issues/11935#issuecomment-2135846313) (among others) for the poor fuzzing results.

renatahodovan:
Minimal qjs repro of the infinite loop that tries to match the empty string non-greedily:

```js
var r = new RegExp("()*?a");
r.exec(",");
```

Execute: `./qjs test.js`

libregexp.c
```diff
@@ -1995,6 +1996,10 @@ static intptr_t lre_exec_backtrack(REExecContext *s, uint8_t **capture,
 
     for(;;) {
         //        printf("top=%p: pc=%d\n", th_list.top, (int)(pc - (bc_buf + RE_HEADER_LEN)));
+        if (++s->backtrack_count > 1000) {
```
saghul, line 1999: IMHO this should likely be a `define` with some sane default value, but overridable at compile time.

saghul (commented):
Left a comment! @chqrlie you probably want to have a look.

bellard:
I think it is better to fix the non greedy matching first. This patch only hides the real problem.

bellard:
The commit `36911f0d` should fix the issue. Do you have other cases where the backtracking limitation workaround is necessary ?

renatahodovan:
@bellard Thanks for the fix, it really fixed the recursion in the previous test case! However, I have another repro:

```js
var r = new RegExp("(.+)+a");
r.exec("VVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVV");
```

VadimZhestikov:
> @bellard Thanks for the fix, it really fixed the recursion in the previous test case! However, I have another repro:
> 
> ```js
> var r = new RegExp("(.+)+a");
> r.exec("VVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVV");
> ```

This example is not about infinite recursion. It is just demonstrate valid regexp behavior for given pattern, for which regexp execution time growing very fast with growing length of input string

renatahodovan:
> This example is not about infinite recursion. It is just demonstrate valid regexp behavior for given pattern, for which regexp execution time growing very fast with growing length of input string

You are right (I think), but I believe we also agree that such inputs are either useless or malicious. Therefore, if we limit the number of recursive calls they trigger, we won't be restricting the actual capabilities of the tool, but we can prevent malicious loops. (I killed the above code after 6 minutes, and who knows how long it would have run. And the input isn't even that long...)

renatahodovan:
Oh, it seems that the comment I replied to has disappeared in the meantime.

bellard:
I see no problem with your example because it really requires a large number of backtracking (try it in another JS engine). It is the same problem as having a JS function which takes a long time to execute.

However, there is a real problem that should be addressed: the interrupt handler registered with JS_SetInterruptHandler() is not called during regexp execution which prevents the user from easily interrupting their execution.

renatahodovan:
@bellard Is there any update on the correct method for registering an interrupt handler during regex execution? Alternatively, could the above patch serve as a temporary solution to address the frequent timeout issues during fuzzing, which are causing early exits and poor performance?

renatahodovan:
@bellard: I think I found a solution: we can create a fuzzer-friendly build, as suggested by the libFuzzer [documentation](https://llvm.org/docs/LibFuzzer.html#fuzzer-friendly-build-mode), using the `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` macro. This way, the backtracking limit will only be applied for fuzzing and will not affect the normal mode operation of the JS engine, while also preventing the fuzz session from continuously timing out.

renatahodovan:
Gentle ping

renatahodovan:
@saghul @bellard what do you think of the new fuzzer-specific approach?

saghul:
The change LGTM (perhaps the name could use some bikesheding) but I have no commit rights here :-)

renatahodovan:
@saghul Thanks for the LGTM! The question about the name of the environment variable is an "easy" one, since it is a constant [set by OSS-Fuzz](https://github.com/google/oss-fuzz/blob/c70e5127f05e2d6a025b6b92191ada6fd36bcf08/infra/base-images/base-clang/Dockerfile#L65) and it is also the officially recommended name by [LLVM](https://llvm.org/docs/LibFuzzer.html#fuzzer-friendly-build-mode) for creating fuzzer friendly build from the target. :-)

bellard:
the issue was fixed by adding an interrupt handling in regexp.

PR #312, opened 2024-05-30
Revert "Add `JS_HasException()` (#265)"

bellard:
This reverts commit db9dbd0a2b6d115c9ef3c0dc37e0c669ef4844e4.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6402,11 +6402,6 @@ JSValue JS_GetException(JSContext *ctx)
     return val;
 }
 
-JS_BOOL JS_HasException(JSContext *ctx)
-{
-    return !JS_IsNull(ctx->rt->current_exception);
-}
-
 static void dbuf_put_leb128(DynBuf *s, uint32_t v)
 {
     uint32_t a;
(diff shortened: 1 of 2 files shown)


bellard:
This patch is not correct because "null" is a valid exception. Currently, the only way to test an exception is to test if JS_EXCEPTION is returned. The content of "current_exception" is undefined otherwise.

kasperisager:
Would it be reasonable to make the contents `current_exception` of defined at all times, such as by initialising it to `JS_NULL`, so one can actually test for pending exceptions?

kasperisager:
That's an awfully good point! Initialising it to `JS_UNINITIALIZED` instead and letting that mean "no pending exception" sounds like a plan, I can put together a PR.

chqrlie:
> That's an awfully good point! Initialising it to `JS_UNINITIALIZED` instead and letting that mean "no pending exception" sounds like a plan, I can put together a PR.

What about `throw undefined;` ?  If any value can be thrown and should be detectable with `JS_HasException`, we should have a boolean in addition to the `current_exception` field.

kasperisager:
@chqrlie Isn't `JS_UNDEFINED` it's own thing?

chqrlie:
> @chqrlie Isn't `JS_UNDEFINED` it's own thing?

Yes, I was confused in my earlier comment. `JS_UNDEFINED` and `JS_UNINITIALIZED` are different types, `JS_UNINITALIZED is used for special `let` semantics and is a perfect solution for the issue at stake here.

PR #313, opened 2024-05-30
Fix `JS_HasException()` when `null` is thrown

kasperisager:
Closes #312


chqrlie (approved):
Using `JS_UNINITIALIZED` is a clever solution to allow for any valid value to be thrown.

saghul:
@chqrlie Shall this be merged?

PR #317, opened 2024-06-09
Fix non-ASCII strings in `JS_NewAtomLen`

Emill:
Fixes the following bug (run in `qjs`):

```javascript
Ãµ = 3;
console.log(õ); // prints 3, should throw ReferenceError
```


bellard:
fixed

PR #334, opened 2024-07-23
Ensure that workers in fuzzers can create their own context

renatahodovan:
Inspired by qjs, a new helper method was added to create the JS context, that can be reused to create context in workers, too.


saghul (approved):

renatahodovan:
@saghul Thanks for your approval! Could someone land it please?

renatahodovan:
@bellard Could you please take a look?

Issue #98, opened 2021-10-27
Worker.parent.onmessage not set to null in test_worker_module.js test


chemwolf6922:
In the test 'test_worker.js', after parent thread set worker.onmesssage to null and 'terminated' the worker, the worker would still be running since its parent.onmessage was not set to null. <br>
This does not matter in this test, since the parent thread will exit soon afterwards. However, if there were any more async operations in the parent thread, the detached worker would always be running in the background. <br>
Thus, please consider setting parent.onmessage to null after the worker posts 'done' to its parent in this test/example.

bellard:
fixed

PR #336, opened 2024-07-27
Fix termination in Worker test

akosthekiss:
Function names are case sensitive, `onMessage` is not the same as `onmessage`.

Related to #98


saghul (approved):

akosthekiss:
@chqrlie As far as I know, the Worker API of QuickJS is based on the Web Worker API (https://github.com/bellard/quickjs/blob/master/doc/quickjs.texi#L813). According to the HTML Living Standard (https://html.spec.whatwg.org/#workers), workers have `onmessage` (all lower case) and `postMessage` (camel case).

PR #339, opened 2024-08-07
Expose `JS_ToBoolean()`, `JS_ToNumber()`, and `JS_ToObject()`

kasperisager:
We're using these for type coercion: https://github.com/holepunchto/libqjs/blob/06c724c571798cb99ec777ff50fe37a68e79b3a0/src/qjs.c#L2925-L2997


kasperisager:
Moved to https://github.com/quickjs-ng/quickjs/pull/824.

PR #353, opened 2024-09-19
Add JS_DumpObjects support

HarlonWang:
To analyze memory leaks and usage more accurately, it would be great to expose `JS_DumpObjects `.
image

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2295,17 +2295,7 @@ void JS_FreeContext(JSContext *ctx)
     JS_DumpShapes(ctx->rt);
 #endif
 #ifdef DUMP_OBJECTS
-    {
-        struct list_head *el;
-        JSGCObjectHeader *p;
-        printf("JSObjects: {\n");
-        JS_DumpObjectHeader(ctx->rt);
-        list_for_each(el, &rt->gc_obj_list) {
-            p = list_entry(el, JSGCObjectHeader, link);
-            JS_DumpGCObject(rt, p);
-        }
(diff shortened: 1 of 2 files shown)


saghul:
It would be nice if the function got a `FILE *f` parameter so it would work with any stream other than stdout too.

HarlonWang:
I also think that's better, but the changes to the original logic will be quite extensive. The following methods will all need adjustments(add a FILE *f parameter).
```c
static __maybe_unused void JS_DumpAtoms(JSRuntime *rt);
static __maybe_unused void JS_DumpString(JSRuntime *rt, const JSString *p);
static __maybe_unused void JS_DumpObjectHeader(JSRuntime *rt);
static __maybe_unused void JS_DumpObject(JSRuntime *rt, JSObject *p);
static __maybe_unused void JS_DumpGCObject(JSRuntime *rt, JSGCObjectHeader *p);
static __maybe_unused void JS_DumpValueShort(JSRuntime *rt, JSValueConst val);
static __maybe_unused void JS_DumpValue(JSContext *ctx, JSValueConst val);
```
I can try making these changes, but I'm not sure if they'll be accepted. 

Do you have any better suggestions? @saghul 

By the way, I often use this method to analyze leaked objects. in my use case, I achieved writing to a file by proxying stdout, so it would be great if this could be officially supported.

saghul:
I'd accept the change in the friendly fork though ;-)

johnjg75dev (approved):

Issue #326, opened 2024-06-28
bf_atof_internal should check `USE_BF_DEC` macro before using bfdec_normalize_and_round


laishere:
When I tried to compile quickjs, I found there's a missing macro check in bf_atof_internal  function.
https://github.com/bellard/quickjs/blob/36911f0d3ab1a4c190a4d5cbe7c2db225a455389/libbf.c#L3110

`bfdec_normalize_and_round` is defined under `USE_BF_DEC` flag, and `bf_atof_internal `  trys to access it without the same flag.

chqrlie:
Thank you for reporting this.  I shall fix this bug shortly.

HarlonWang:
I think this issue hasn't been fixed yet. I made a simple fix, I'm not sure if it's the same as your fix. You can take a look https://github.com/bellard/quickjs/pull/354 for details. @chqrlie

bellard:
libbf.c was removed.

Issue #241, opened 2024-02-17
Undefined symbol `_bfdec_normalize_and_round`, when compiling without `CONFIG_BIGNUM`, since 94010ed


past-due:
In 94010edb5794b06e122cfb9ea6ed9c6be9db6c03, the defines for `USE_BF_DEC` and `USE_FFT_MUL` were placed inside an `#ifdef CONFIG_BIGNUM`:

https://github.com/bellard/quickjs/blob/94010edb5794b06e122cfb9ea6ed9c6be9db6c03/libbf.c#L40-L45

This causes an undefined symbol error when compiling without `CONFIG_BIGNUM` defined, due to `_bf_atof_internal`'s use of `_bfdec_normalize_and_round`.

```
ld: Undefined symbols:
  _bfdec_normalize_and_round, referenced from:
      _bf_atof_internal in libquickjs.a[3](libbf.o)
```

(Example: Try compiling QuickJS without optimizations and without CONFIG_BIGNUM defined.)

It appears that there are only two uses of `_bf_atof_internal` when `CONFIG_BIGNUM` is undefined:

https://github.com/bellard/quickjs/blob/3bb2ca360e994c93e9bf273a685c061d5e0b58f4/libbf.c#L3159-L3172

Both of these pass `FALSE` to the `BOOL is_dec` parameter of `_bf_atof_internal`, and so `_bfdec_normalize_and_round` should never actually be called when `CONFIG_BIGNUM` is undefined:

https://github.com/bellard/quickjs/blob/3bb2ca360e994c93e9bf273a685c061d5e0b58f4/libbf.c#L3105-L3108

----

So it seems like `_bf_atof_internal` could use an `#ifdef USE_BF_DEC` check around this single use of `bfdec_normalize_and_round`. For example (not sure if this is the best / desired approach):

```diff
      if (is_dec) {
+ #ifdef USE_BF_DEC
          a->expn = expn + int_len;
          a->sign = is_neg;
          ret = bfdec_normalize_and_round((bfdec_t *)a, prec, flags);
+ #else
+         // bfdec_normalize_and_round not available without USE_BF_DEC
+         // if (!radix_bits)
+         //    bf_delete(a); // needed?
+         bf_set_nan(r);
+         ret = BF_ST_MEM_ERROR;
+ #endif
      } else if (radix_bits) {
```

@chqrlie : Any thoughts / suggestions?

chqrlie:
[...]

PR #354, opened 2024-09-19
Fix compiling with undefined to `bfdec_normalize_and_round`

HarlonWang:
A compilation error occurs when without CONFIG_BIGNUM.
```text
quickjs/libbf.c:3110: error: undefined reference to 'bfdec_normalize_and_round'
```
Because there's no conditional handling here.
```c
/* enable it to check the multiplication result */
//#define USE_MUL_CHECK
#ifdef CONFIG_BIGNUM
/* enable it to use FFT/NTT multiplication */
#define USE_FFT_MUL
/* enable decimal floating point support */
#define USE_BF_DEC
#endif

static int bf_atof_internal(...){
    if (is_dec) {
        a->expn = expn + int_len;
        a->sign = is_neg;
        // we need to check if USE_BF_DEC is enabled.
        ret = bfdec_normalize_and_round((bfdec_t *)a, prec, flags);
    } else if (radix_bits) {
        ...
    }
   ...
```

I made a simple fix.
```c
#ifdef USE_BF_DEC
    if (is_dec) {
        a->expn = expn + int_len;
        a->sign = is_neg;
        ret = bfdec_normalize_and_round((bfdec_t *)a, prec, flags);
    } else
#endif
    if (radix_bits) {
```

Related issues:
https://github.com/bellard/quickjs/issues/241
https://github.com/bellard/quickjs/issues/326

### DIFF
--- a/libbf.c
+++ b/libbf.c
@@ -3104,11 +3104,14 @@ static int bf_atof_internal(bf_t *r, slimb_t *pexponent,
         if (exp_is_neg)
             expn = -expn;
     }
+#ifdef USE_BF_DEC
     if (is_dec) {
         a->expn = expn + int_len;
         a->sign = is_neg;
         ret = bfdec_normalize_and_round((bfdec_t *)a, prec, flags);
-    } else if (radix_bits) {
+    } else
+#endif
+    if (radix_bits) {
         /* XXX: may overflow */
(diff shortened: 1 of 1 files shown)


bellard:
fixed.

PR #356, opened 2024-09-25
skip verification of class ID in JS_GetOpaque if class_id is 0

bendmorris:
When using QuickJS from C++ I'd like to be able to get the opaque pointer back without necessarily knowing the class ID of the value.

As an example, I have a Parent class in C++ and multiple classes that extend it: Child1, Child2..., each with their own wrapping JS class. I want to create a JS function which can take any of these child classes, cast the opaque value to a `Parent *` and call methods on that.

Since 0 is not a valid class ID, using it to disable verification works well.


maarrat (approved):

bellard:
You can use GetAnyOpaque() instead.

Issue #183, opened 2023-06-04
Profinling Javascript with quickjs ?


mingodad:
Does someone managed to profile Javascript code with quikjs ?
Ideally `qjs` would have an option to save/output a profile report of the running script, something like:
```
qjs -p myscript.js myscript_arg
```
I'm looking at the source code to try find the ideal point to install hooks that allow such functionality, any help is welcome !

Cheers !

mingodad:
I've got an initial implementation of it that outputs something like:
```
[1	2	1	103	<eval>]
[16589	1	1	13314514	load_model]
[16248	1	1	2	glp_mpl_free_wksp]
[16315	1	1	13304350	glp_mpl_generate]
[13023	1	1	13304347	mpl_generate]
[12441	1	1	254	mpl_internal_build_problem]
[12424	1	1	13304079	mpl_internal_generate_model]
[16243	1	1	229	glp_mpl_alloc_wksp]
[12763	1	1	226	mpl_initialize]
[746	1	1	211	rng_create_rand]
[491	1	1	4	glp_version]
[14860	1	1	1	mpl_get_num_cols]
[14841	1	1	1	mpl_get_num_rows]
[16260	1	1	9708	glp_mpl_read_model]
[12887	1	1	9705	mpl_read_model]
[12889	1	1	32	<null>]
[5720	1	1	11	mpl_internal_end_statement]
[6994	1	1	981	mpl_internal_data_section]
[6865	1	1	173	mpl_internal_parameter_data]
[6553	1	1	50	mpl_internal_plain_format]
[6506	1	1	47	mpl_internal_read_value]
[6016	1	1	24	mpl_internal_read_number]
[6462	1	1	1	mpl_internal_select_parameter]
[6326	1	1	791	mpl_internal_set_data]
[6155	20	2	79	mpl_internal_simple_format]
[5935	20	2	2	mpl_internal_slice_arity]
[5953	2	1	2	mpl_internal_fake_slice]
[6118	1	1	1	mpl_internal_select_set]
[5964	2	1	20	mpl_internal_delete_slice]
[6006	9	1	7	mpl_internal_is_literal]
[5991	95	10	15	mpl_internal_is_symbol]
[6029	40	4	49	mpl_internal_read_symbol]
[5981	41	5	7	mpl_internal_is_number]
[5892	2	1	1	mpl_internal_expand_slice]
[12363	1	1	22	mpl_internal_alloc_content]
[...]

PR #371, opened 2024-11-23
Implements Javascript profiling support.

rubenlg:
This pull request addresses #183 and builds upon the ideas introduced in #184, taking a slightly different approach to enhance flexibility and functionality:
- Profiling configuration is achieved by providing callback functions and optional opaque data. QuickJS will invoke these callbacks with relevant atoms for the function name and filename, along with the provided opaque data (e.g., a file handle for writing output). If possible, the function name includes the constructor of the "this" object, adding context. Unlike #184, which logs stats directly to stderr, this implementation offers greater flexibility by enabling client applications to massage the data in different ways (computing aggregated statistics, logging the raw events, etc).
- In #184, logging occurred in the bytecode destructor, which often led to function name atoms being collected prematurely, resulting in many functions being logged as `<:isGone:>`. This implementation resolves that by reporting profiling data at function entry and exit, ensuring the function name remains valid. Additionally, this approach includes context for functions (e.g., the constructor of the `this` object, akin to V8 stack traces for disambiguation) and flags anonymous functions explicitly.
- The qjs binary demonstrates an example usage of the profiling API by producing output in a format compatible with Chrome DevTools, chrome://tracing and tools like [Perfetto](https://ui.perfetto.dev/). By default, every call is logged, though the API (and qjs) supports configurable sampling. Here is an example of running `./qjs -p profile.json examples/pi_bigint.js 20` and then opening profile.json in the Chrome Dev Tools:
image

Please note that I'm not super familiar with QuickJS yet. I downloaded the code this week and have been playing with it only for a couple of days. Let me know if there are better ways to do some of the things I'm doing here, or if there is a coding style guide this change should adhere to.

Thanks!

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -289,12 +289,49 @@ void help(void)
 #endif
            "-T  --trace        trace memory allocation\n"
            "-d  --dump         dump the memory usage stats\n"
+#ifdef CONFIG_PROFILE_CALLS
+           "-p  --profile FILE dump the profiling stats to FILENAME in a format that can be open in chromium devtools\n"
+           "    --profile-sampling n collect function calls every 1/n times. Defaults to 1 (i.e. every call)\n"
+#endif
            "    --memory-limit n       limit the memory usage to 'n' bytes\n"
            "    --stack-size n         limit the stack size to 'n' bytes\n"
            "    --unhandled-rejection  dump unhandled promise rejections\n"
            "-q  --quit         just instantiate the interpreter and quit\n");
     exit(1);
 }
+
+#ifdef CONFIG_PROFILE_CALLS
+void profile_function_start(JSContext *ctx, JSAtom func, JSAtom filename, void *opaque) {
+    const char* func_str = JS_AtomToCString(ctx, func);
+    const char *func_str2 = func_str[0] ? func_str : "<anonymous>";
+    FILE *logfile = (FILE *)opaque;
+    // Format documented here:
+    // https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.yr4qxyxotyw
+    fprintf(logfile, "{"
+            "\"name\": \"%s\","
+            "\"cat\": \"js\","
+            "\"ph\": \"B\","
+            "\"ts\": %ld,"
+            "\"pid\": 1,"
+            "\"tid\": 1"
+        "},\n", func_str2, clock());
+    JS_FreeCString(ctx, func_str);
+}
+void profile_function_end(JSContext *ctx, JSAtom func, JSAtom filename, void *opaque) {
+    const char* func_str = JS_AtomToCString(ctx, func);
+    const char *func_str2 = func_str[0] ? func_str : "<anonymous>";
+    FILE *logfile = (FILE *)opaque;
+    fprintf(logfile, "{"
+            "\"name\": \"%s\","
+            "\"cat\": \"js\","
+            "\"ph\": \"E\","
+            "\"ts\": %ld,"
+            "\"pid\": 1,"
+            "\"tid\": 1"
+        "},\n", func_str2, clock());
+    JS_FreeCString(ctx, func_str);
+}
+#endif
 
 int main(int argc, char **argv)
 {
@@ -306,6 +343,10 @@ int main(int argc, char **argv)
     int interactive = 0;
     int dump_memory = 0;
     int trace_memory = 0;
+#ifdef CONFIG_PROFILE_CALLS
+    FILE *profile_file = NULL;
+    uint32_t profile_sampling = 1;
+#endif
     int empty_run = 0;
     int module = -1;
     int load_std = 0;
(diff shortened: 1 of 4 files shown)


rubenlg:
By the way, the current implementation in the `qjs` program is super naive, issuing an `fwrite` on every function start/end, which is expensive and likely skewing the measurements (especially since `fwrite` does buffering and some calls will get an arbitrary penalty when hitting the disk).

If `qjs` is used in production environments and not just as an example of how to use the quickjs library (I don't know, please tell me), I could change the implementation to collect the events in memory and write them at the end instead. I've already written that code for my own application so it's just a matter of porting it from C++ to C.

2bam:
This is excellent, thanks Rubén. You published it just in time for something I needed.
¡Moitas grazas!

I ran into some issues probably due to embedding quickjs through [quickjspp](https://github.com/ftk/quickjspp) instead of running in qjs.

### Issues

⚠️ The most important one: I'm getting garbled names coming from random location pointers, which crash the visual tools (specially '\n's).

I worked around that sanitizing for the time being because I didn't have time to dig into the qjs codebase and figure out what went wrong. Definitely not great for overhead but still helped me a lot.

 ```
void sanitize(const char* in, char* out, int n) {
    n--;
    for (int i = 0; *in && i < n; i++) {
        *(out++) = isprint(*in) ? *in : 'X';
        in++;
    }
    *out = 0;
}
//...
char ok[128];
sanitize(func_str2, ok, 128);
```

### Tools

First the good news, [ui.perfetto.dev](https://ui.perfetto.dev/) works perfectly well, even with a lot of data and some OOM crashes.

Chrome Tools' profiler on the other hand is **really bad** at loading the profile. I’m getting: `Malformed timeline data: SyntaxError: Unexpected end of JSON input` even when it's definitely not malformed and still loads partially, approximately only the first 4 seconds of a 120 seconds run at a 1.5GB file. And forget about loading a file >2GB due to 32 (31?) bit limits.

### Patching quickjspp

As quickjspp integrates quickjs manually without subtrees/submodules I'll leave this script to patch and build quickjspp for reference in case anyone wants to try this great profiler out and is in the same situation as I am.

```
#!/bin/bash

mkdir -p temp
git clone https://github.com/rubenlg/quickjs.git temp/quickjs-profiling
cd temp/quickjs-profiling/
git checkout profiling
git diff master >../quickjs-profiling.patch

cd ../../quickjspp/
git apply -v ../temp/quickjs-profiling.patch --directory=quickjs --reject --whitespace=fix
if [ ! -f CMakeLists.txt.profbak ]; then
    cp CMakeLists.txt CMakeLists.txt.profbak
    echo "set_property(DIRECTORY quickjs PROPERTY COMPILE_DEFINITIONS CONFIG_PROFILE_CALLS=1)" >>CMakeLists.txt
fi
mkdir -p build
cd build
cmake ..
cmake --build .
```

**Comment by rubenlg** [comment]
I also had problems with loading large profile traces in Chrome Profiler. I ended up creating a tree of calls in memory and pruning the fast branches (those that take less than 0.1 ms to execute) before writing the file. This made the file much smaller and more manageable. I didn't do any of that for the qjs binary in this PR, though.

The garbled names could be caused by how quickjspp maps the C++ API to quickjs. I can't reproduce that myself. When I expose C++ functions and classes to quickjs, I control the names I give them, and that's what I get back in the profiles. I also do multiple jumps between C++ and Javascript and yet the profiles are clean.

**Comment by 2bam** [comment]
I had some time today and I figured out the issue, `get_full_func_name` can return `JS_ATOM_NULL` (`0`), passed on to `JS_AtomToCString` which doesn't do a null check, it even calls JS_DupValue making quickjs refcount _...something_ 😱

You might want to update your `profile_function_start/end` to check for that.

```
static JSAtom get_full_func_name(JSContext *ctx, JSValueConst func, JSValueConst this_obj)
{
    JSAtom result_atom;
    const char *func_str = get_func_name(ctx, func);
    if (!func_str || func_str[0] == '\0') {
        JS_FreeCString(ctx, func_str);
        return JS_ATOM_NULL;
    }
    //...
```

**Comment by rubenlg** [comment]
Thanks @2bam I'll take a look after vacation.

File: qjs.c [hunk 1]
```diff
@@ -289,13 +289,50 @@ void help(void)
...
            "    --memory-limit n       limit the memory usage to 'n' bytes\n"
            "    --stack-size n         limit the stack size to 'n' bytes\n"
            "    --unhandled-rejection  dump unhandled promise rejections\n"
            "-q  --quit         just instantiate the interpreter and quit\n");
     exit(1);
 }
 
+#ifdef CONFIG_PROFILE_CALLS
+void profile_function_start(JSContext *ctx, JSAtom func, JSAtom filename, void *opaque) {
+    const char* func_str = JS_AtomToCString(ctx, func);
```
2bam, line 305: Here the `func` parameter can receive a JS_ATOM_NULL (passed on from a `get_full_func_name` returned value), and `JS_AtomToCString` will not do a null check causing unexpected behavior.
Checking before (like something similar to the line below) will solve the issue.

Same goes for `profile_function_end`.

rubenlg: Thanks! Should be addressed in the latest commit.

Issue #314, opened 2024-06-01
"invalid escape sequence in regular expression"


jclab-joseph:
An error occurred while parsing regex.

It doesn't seem to recognize `[a\-b]`.

### SOURCE CODE

https://github.com/jsdom/xml-name-validator/blob/main/lib/xml-name-validator.js

```
var validateNames = {};
var xmlNameValidator = {};
xmlNameValidator.name = (potentialName) => {
  return /^[:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u{10000}-\u{EFFFF}][:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\u{10000}-\u{EFFFF}\-.0-9\u00B7\u0300-\u036F\u203F-\u2040]*$/u.test(potentialName);
};
xmlNameValidator.qname = (potentialQname) => {
[...]

PR #373, opened 2024-12-13
regexp: Allow [\-] even in unicode mode

parnikkapore:
Hi! This is a straightforward fix for https://github.com/bellard/quickjs/issues/314.

I couldn't find information on what the correct behavior is, but this corresponds to the other JS engines I tried (Chrome + FF + Node).

Code style comments also welcome. There's probably a more preferable way to do this.

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -700,8 +700,10 @@ static int get_class_atom(REParseState *s, CharRange *cr,
             if (ret >= 0) {
                 c = ret;
             } else {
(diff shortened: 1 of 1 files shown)


parnikkapore:
...or just take https://github.com/quickjs-ng/quickjs/commit/f406d6f78c0dfe589236b8a19e91e0e403009805 from the friendly fork I didn't know existed. That works too.

bellard:
equivalent fix commited.

parnikkapore:
Ooh, that does look cleaner. Welcome back!

Issue #363, opened 2024-10-30
No error showed when error threw with `--module` flag


zhuscat:
QuickJS Version: 2024-01-13

```js
/* fib module */
export function fib(n)
{
    if (n <= 0)
        return 0;
    else if (n == 1)
        return 1;
    else
        return fib(n - 1) + fib(n - 2);
}
```

```js
/* example of JS module */

import { fib } from "./fib_module.js";

// here I throw an Error
throw new Error('test error');

console.log("Hello World");
console.log("fib(10)=", fib(10));
```

When I execute `qjs ./examples/hello_module.js`, no error showed

navrocky:
I confirm this error for a version 2024-01-13. After rolling back to version 2023-12-09 the errors started to appear.

I am using the QuickJS API,  not the CLI. When I call `JS_Eval` with a broken script and `JS_EVAL_TYPE_GLOBAL` flag then it returns JS exception (it's normal):
```
JS_Eval(ctx, script.c_str(), script.size(), file.c_str(), JS_EVAL_TYPE_GLOBAL);
```

But if I  `JS_EVAL_TYPE_MODULE` flag then JS_Eval silently terminates without any errors or exceptions and returns JSValue with JS_TAG_MODULE: 

```
JS_Eval(ctx, script.c_str(), script.size(), file.c_str(), JS_EVAL_TYPE_MODULE);
```

Logging insertions into the script shows that execution flow aborted right at the point of  the error.

In version 2023-12-09 everything works with modules as expected.

bnoordhuis:
Does it work when you pass the JS_Eval result to JS_EvalFunction? And in turn pass _that_ result to js_std_await?

I didn't test with the code from this repo but in quickjs-ng it works as expected.

XuJiandong:
a very minimal reproducible case:

```
❯ cat throw.mjs

throw "hi";

❯ qjs throw.mjs; echo $?
0
```

saghul:
The CLI won't exit with an error in async functions. We fixed that in NG: https://github.com/quickjs-ng/quickjs/pull/514

soulofmischief:
> The CLI won't exit with an error in async functions. We fixed that in NG: [quickjs-ng/quickjs#514](https://github.com/quickjs-ng/quickjs/pull/514)

Thank you, I upstreamed part of that into a PR for this repository https://github.com/bellard/quickjs/pull/386

soulofmischief:
[...]

PR #386, opened 2025-03-11
Display errors in async methods and at top level

soulofmischief:
This fixes a regression caused by the most recent release.

I'm unfamiliar with the project, I just upstreamed https://github.com/quickjs-ng/quickjs/pull/811/commits/77332f2eb16ad6608dfafcec3486d541bc848350 so this patch needs review. 

Tested manually locally and it seems to work. I didn't upstream the test file because it seems to be missing from quickjs-ng's latest build. I was unsure if it should also be upstreamed or written from scratch to be more idiomatic with this project.

Should fix https://github.com/bellard/quickjs/issues/363 and maybe others

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -25,6 +25,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <inttypes.h>
 #include <string.h>
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


bnoordhuis:
interrupt-test.c is called api-test.c in quickjs-ng now but contains more tests that in all likelihood won't pass with bellard/master.

When cherry-picking commits, it's considered good style to retain the original Author. The way you did it now makes it look like you wrote that code.

soulofmischief:
That's a good point, I didn't think to credit the original author since it was a small patch and there were discrepancies which meant I had to modify the patches to work with the upstream, but I have no problem crediting laishere. Thanks for pointing that out.

soulofmischief:
Closing this PR anyway as Fabrice says the issue is resolved in the most recent version

PR #391, opened 2025-03-18
Fix multiarch CI builds

nickva:
Try to fix https://github.com/bellard/quickjs/issues/390


nickva:
Screenshot 2025-03-18 at 7 56 00 PM

PR #405, opened 2025-05-08
os.socket API proposal

yne:
This PR aim to start a discussion about the future socket API.

My goal is use cosmo+QuickJS/HTTP server to offer a portable cross-platform GUI.

I took inspiration from other std syscall wrapper to do this proposal

- `fd  = std.socket(domain=AF_INET, type=SOCK_STREAM, protocol=0)`
- `err = std.bind(sock, {addr:XXX,port:XXX})`
- `err = std.listen(sock)`
- `fd  = std.accept(sock)`
- `err = std.connect(sock, {addr:XXX,port:XXX})`

I'm confident about the return value API, but I'm open to feedback about:
- allowing implicit args (`socket()` currently return a TCP fd, `listen()` default to a backlog of 10)
- shall we use a specific `SockAddr()` constructor (with .parse() and .toString()) instead of the current plain `{address:number,port:number}` object used in `bind()` and `connect()`
- are socket more `os` related than `std`
- remove the `SO_REUSEADDR` I added, and add a setsockopt wrapper (but just for this usecase ?)

Once agreed, I'll add required unit tests/docs ...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,6 +247,8 @@ HOST_LIBS=-lm -ldl -lpthread
 LIBS=-lm -lpthread
 ifndef CONFIG_WIN32
 LIBS+=-ldl
+else
+LIBS+=-lws2_32
 endif
 LIBS+=$(EXTRA_LIBS)
 
--- a/doc/globals.d.ts
+++ b/doc/globals.d.ts
@@ -1,0 +1,14 @@
+/** Command line arguments where argument[0] is the JS script name. */
+declare const scriptArgs: string[];
+/** Print args separated by spaces and a trailing newline. */
+declare function print(...args: any[]): void;
+
+declare const console: {
+	/** Print args separated by spaces and a trailing newline. */
+	log: typeof print
+};
+
+interface ImportMeta {
+	url: string;
+	main: boolean;
+}
--- a/doc/os.d.ts
+++ b/doc/os.d.ts
@@ -1,0 +1,290 @@
+/**
+ * Provides Operating System specific functions:
+ * file access, sockets, signals, timers, async I/O, workers (threads)
+ */
+declare module "os" {
+	type Success = 0;
+	type Errno = number;
+	type NegativeErrno = number;
+	type Result<T> = T | NegativeErrno;
+	type ExitStatus = number;
+	type WaitStatus = number;
+	type Branded<T, B> = T & { __brand: B }; // Prevent interchangeable primitives (e.g. number)
+	type OpenOption = Branded<number, "OpenOption">;
+	type SocketFamily = Branded<number, "SocketFamily">;
+	type SocketType = Branded<number, "SocketType">;
+	type SocketOpt = Branded<number, "SocketOpt">;
+	type SocketShutOpt = Branded<number, "SocketShutOpt">;
+	type PollOpt = Branded<number, "PollOpt">;
+	type PollRet = Branded<number, "PollRet">;
+	type WaitpidOpt = Branded<number, "WaitpidOpt">;
+	type FileDescriptor = Branded<number, "FileDescriptor">;
+	type Signal = Branded<number, "Signal">;
+	type StatMode = Branded<number, "StatMode">;
+	type Pid = Branded<number, "Pid">;
+	type TimerHandle = Branded<unknown, "TimerHandle">;
+	type Callback = () => void;
+	type Platform = "linux" | "darwin" | "win32" | "js";
+	type WorkerMessage = any;
+
+	interface ExecOptions {
+		/**
+		 * Boolean (default = `true`). If `true`, wait until the process is
+		 * terminated. In this case, exec return the exit code if positive or the
+		 * negated signal number if the process was interrupted by a signal. If
+		 * false, do not block and return the process id of the child.
+		 */
+		block?: boolean;
+		/** Is the file searched in the PATH @default true. */
+		usePath?: boolean;
+		/** Set the file to be executed @default args[0] */
+		file?: string,
+		/** The working directory of the new process */
+		cwd?: string,
+		/** Handler for stdin */
+		stdin?: FileDescriptor,
+		/** Handler for stdout */
+		stdout?: FileDescriptor,
+		/** Handler for stderr */
+		stderr?: FileDescriptor,
+		/** set the process environment from the object key-value pairs.
+		 * If unset, it will inherit them from current process. */
+		env?: { [key: string]: string },
+		/** Process uid with `setuid` */
+		uid?: number,
+		/** Process gid with `setgid` */
+		gid?: number,
+	}
+	type ExecNonBlockingOptions = ExecOptions & { block?: false };
+	type ExecBlockingOptions = ExecOptions & { block: true };
+
+	interface Stat {
+		dev: number;
+		ino: number;
+		mode: StatMode;
+		nlink: number;
+		uid: number;
+		gid: number;
+		rdev: number;
+		size: number;
+		blocks: number;
+		/** milliseconds since 1970 */
+		atime: number;
+		/** milliseconds since 1970 */
+		mtime: number;
+		/** milliseconds since 1970 */
+		ctime: number;
+	}
+	interface SocketAddr {
+		/** @example 80 */
+		port: number;
+		/** @example "8.8.8.8" or "::1" */
+		addr: string;
+		/** @default AF_INET */
+		family: SocketFamily;
+	}
+	interface SocketAddrInfo extends SocketAddr {
+		socktype: SocketType;
+	}
+	interface HintAddr {
+		/** @example 80 or "http" */
+		service?: number | string;
+		/** @default any */
+		family?: SocketFamily;
+		/** @default any */
+		socktype?: SocketType;
+	}
+	/* POSIX constants (Windows-specific) */
+	export const O_TEXT: OpenOption;
+	/* POSIX constants */
+	export const O_APPEND: OpenOption;
+	export const O_CREAT: OpenOption;
+	export const O_EXCL: OpenOption;
+	export const O_RDONLY: OpenOption;
+	export const O_RDWR: OpenOption;
+	export const O_TRUNC: OpenOption;
+	export const O_WRONLY: OpenOption;
+	export const S_IFBLK: StatMode;
+	export const S_IFCHR: StatMode;
+	export const S_IFDIR: StatMode;
+	export const S_IFIFO: StatMode;
+	export const S_IFLNK: StatMode;
+	export const S_IFMT: StatMode;
+	export const S_IFREG: StatMode;
+	export const S_IFSOCK: StatMode;
+	export const S_ISGID: StatMode;
+	export const S_ISUID: StatMode;
+	export const SIGABRT: Signal;
+	export const SIGALRM: Signal;
+	export const SIGCHLD: Signal;
+	export const SIGCONT: Signal;
+	export const SIGFPE: Signal;
+	export const SIGILL: Signal;
+	export const SIGINT: Signal;
+	export const SIGPIPE: Signal;
+	export const SIGQUIT: Signal;
+	export const SIGSEGV: Signal;
+	export const SIGSTOP: Signal;
+	export const SIGTERM: Signal;
+	export const SIGTSTP: Signal;
+	export const SIGTTIN: Signal;
+	export const SIGTTOU: Signal;
+	export const SIGUSR1: Signal;
+	export const SIGUSR2: Signal;
+	export const WNOHANG: WaitpidOpt;
+	export const AF_INET: SocketFamily;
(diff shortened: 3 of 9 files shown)


bellard:
I am interested by the feature. Here are some suggestions:

- the functions should be in "os" instead of "std"
- no implicit args for socket()
- implicit arg for listen is acceptable
- use a plain object for sockaddr. The "family" field is optional (AF_INET by default). "address" should be renamed to "addr" and should be a string containing the IPv4 or IPv6 address. AF_INET6 should be supported too
- remove SO_REUSEADDR and add setsockopt() and getsockopt()
- For better portability it is useful to add recv, recvfrom, send, sendto and shutdown
- In your example, it is better to avoid using fdopen() on sockets ('FILE *' are designed to work on files)
- a wrapper to getaddrinfo() would be needed (take a hostname as input, returns a sockaddr object)
- for win32 compilation without cosmolibc some specific support might be needed if you want os.write/os.read/os.close to work with socket handles

yne:
Okay here is an update.

> * the functions should be in "os" instead of "std"

agreed, now moved to os

> * no implicit args for socket()

done, socket now use 2 args (family, type)

> * implicit arg for listen is acceptable

okay, I kept backlog=10

> * use a plain object for sockaddr. The "family" field is optional (AF_INET by default). "address" should be renamed to "addr" and should be a string containing the IPv4 or IPv6 address. AF_INET6 should be supported too

done, now use `{ family: number, port: number, addr: string | null }`

> * remove SO_REUSEADDR and add setsockopt() and getsockopt()

done, `os.setsockopt(sock_srv, os.SO_REUSEADDR, new Uint32Array([1]).buffer)` see example/http_server.js

> * For better portability it is useful to add recv, recvfrom, send, sendto and shutdown

understood, I'll do that, but shall I re-use existing js_os_read_write function and add 4 magics to handle the recv/recvto/send/sendto cases ?
or do you prefere a separate function ?

> * In your example, it is better to avoid using fdopen() on sockets ('FILE *' are designed to work on files)

hum, okay, I've moved to only os.* calls
I've seen cases on other project where fprintf where used on socket (e.g. for an IRC client)
Out of curiosity, is it a non-UNIX limitation ?

> * a wrapper to getaddrinfo() would be needed (take a hostname as input, returns a sockaddr object)

done, I've added one, but I think It will need some &hint values as it return too much results (See: example/http_client.js that need a .filter() to work)

> * for win32 compilation without cosmolibc some specific support might be needed if you want os.write/os.read/os.close to work with socket handles

Sure, I'll try if I get my hands on a windows machine

yne:
> if you want to keep the socket api low-level

For now yes, because all `os` IO function are made that way (synchronous).

Note that existing `os.setReadHandler()`, `os.setWriteHandler()` *might* offer an event-based API that you could wrap as async/generator in JS (I did not tested it myself).

Adding *Async variant to all `os` IO function could be done in a separate PR.

> then it would be great to provide the options to make the sockets non-blocking:
> ```
>     fcntl(sockfd, F_SETFL, fcntl(sockfd, F_GETFL, 0) | O_NONBLOCK);
> ```

Sure, I can add `fcntl()` + `EWOULDBLOCK`/`EAGAIN` support to enable pooling-style of socket;

ceedriic:
> > if you want to keep the socket api low-level
> 
> For now yes, because all `os` IO function are made that way (synchronous).

Kind of true, but filesystem functions are perfectly usable in a synchronous way, whereas a socket API with only synchronous functions is little more than a toy API.

Also look at os.sleepAsync()

> 
> Note that existing `os.setReadHandler()`, `os.setWriteHandler()` offer an callback API for send/recv.
> 

Oh, sorry, missed that. Well, then I think it covers all my use cases (together with your proposed SOCK_NONBLOCK).

If you can also use os.setReadHandler() after listen() and os.setWriteHandler() after a connect() call, then I think everything is covered. Have you tested that?

> Adding *Async variant to all `os` IO function could be done in a separate PR.

> 
> > then it would be great to provide the options to make the sockets non-blocking:
> > ```
> >     fcntl(sockfd, F_SETFL, fcntl(sockfd, F_GETFL, 0) | O_NONBLOCK);
> > ```
> 
> agree, we could either go for:
> 
> * a plain API: add `fcntl()` + `EWOULDBLOCK`/`EAGAIN` support
> * a linux-ish API : allow `SOCK_NONBLOCK` flag at socket() creation which will do the `fcntl()` under the hood to work on all OS.
> 
> I have a preference for the last one but I'll let @bellard have the last word.

I also like SOCK_NONBLOCK fwiw. It seems supported by all the BSDs as well as Linux.

> 
> I'll also add a `poll()` wrapper
> 

I don't think it's the right approach, because once again your poll() function will be synchronous and block other events.
It needs to be integrated in the main quicks_libc.c select() loop.

But in any case, this is moot because if SOCK_NONBLOCK can be set to a socket FD and os.setReadHandler() and os.setWriteHandler() can be attached to sockets FDs, then I believe everything can be built from these primitives.

A connectAsync() that returns a Promise like sleepAsync() would certainly be nice, but as you said, that can be added later.

ceedriic:
Similarly, I guess we want to be able to write something like

```
newSocket = os.acceptAsync(sockfd);
```
or
```
os.acceptAsync(sockfd).then((newSocket) => {
    // do something with newSocket
})

```
So we need something like:

1. make sure the socket is SOCK_NONBLOCK (or put it in that state temporarily with fcntl())
2. call accept(fd, address), then there is 3 possible return values:
2.1) if 0 or positive, a new socket is already available and the promise can be successfully resolved.
2.2) if value is an error other then EWOULDBLOCK/EAGAIN then call the promise error handler, the accept() call has already failed.
2.3) if it is EWOULDBLOCK/EAGAIN, it means we need to wait for a new connection to be accepted. In this case, you need to register the listening socket for read (like os.setReadHandler) on the quickjs event loop.
when the quickjs event loop reports that the socket is signaled for read, call again accept() and call the promise success or error handler based on accept() return value.

bellard:
Here some remarks:

- js_os_poll() should not be renamed.
- I don't think that adding poll() is relevant to this patch. setReadHandler() and setWriteHandler() should suffice.
- async functions can be implemented in Javascript with setReadHandler() and set setWriteHandler(). If not, then there is a problem in the API. You can provide examples in your HTTP client and server.
- If JS_IsException() returns TRUE, then the calling function  must return an exception. This is not the case e.g. in JS_toSockaddrStruct. Moreover, all return values of JS API must be tested and handled.
- using SOCK_NONBLOCK in second argument of socket() is a good idea to set the O_NONBLOCK flag. Another possibility is to open the socket in non blocking mode by default and to add a SOCK_BLOCK flag to make it non blocking.
- only os.send() and os.recv() seem necessary with optional flags and sockaddr parameters.
- implementing IPv6 is necessary in order to have an example of another sockaddr type.

ceedriic:
> Here some remarks:
> * async functions can be implemented in Javascript with setReadHandler() and set setWriteHandler(). If not, then there is a problem in the API. You can provide examples in your HTTP client and server.

@yne: if you want, after you provide the nonblocking sockets, I can write the async wrappers in Javascript, let me know.

yne:
* poll() API removed
* reverted js_os_poll to be the event loop pooling
* more JS_IsException() catched
* IPv6 is on my todolist (along with windows)
* socket is now non blocking mode by default, unless SOCK_BLOCK flag is given
* Fixed os.send() and os.recv() optional parameters

> * async functions can be implemented in Javascript with setReadHandler()/setWriteHandler().

I would prefere to a native async socket API over a setReadHandler/setWriteHandler wrapper that everybody will have to copy. I'm currently trying to implement this

```ts
bind    (sockfd: FileDescriptor, addr: SocketAddr   ): Result<Success>;
listen  (sockfd: FileDescriptor, backlog?: number   ): Result<Success>;
shutdown(sockfd: FileDescriptor, type: SocketShutOpt): Result<Success>;
connect (sockfd: FileDescriptor, addr: SocketAddr   ): Promise<Result<Success>>;
accept  (sockfd: FileDescriptor                     ): Promise<[remotefd: FileDescriptor, remoteaddr: SocketAddr]>;
recv    (sockfd: FileDescriptor, buffer: ArrayBuffer, length?: number                  ): Promise<Result<number>>;
send    (sockfd: FileDescriptor, buffer: ArrayBuffer, length?: number                  ): Promise<Result<number>>;
recvfrom(sockfd: FileDescriptor, buffer: ArrayBuffer, length?: number                  ): Promise<[total: Result<number>, from: SocketAddr]>;
sendto  (sockfd: FileDescriptor, buffer: ArrayBuffer, length : number, addr: SocketAddr): Promise<Result<number>>;
//with type Result<T> = T | NegativeErrno;
```

But the changes are heavier than expected. I'll update when (if) I succeed

@ceedriic non-blocking socket (via O_NONBLOCK flag) are available since https://github.com/bellard/quickjs/commit/c0877189964447064df614aebffeb3105b747f30 if you want to try

yne:
I'm almost finished and it's a joy to play with this API

CHANGES:
- argv order change in `sendto` to keep the last `length?` optional
- bind/connect/sendto `addr` argument rely on `sockfd` family if addr.family is undefined

```ts
//sync calls
socket     (family: SocketFamily, type: SocketType): Result<FileDescriptor>;
getsockopt (sockfd: FileDescriptor, name: SocketOpt, data: ArrayBuffer): Result<Success>;
setsockopt (sockfd: FileDescriptor, name: SocketOpt, data: ArrayBuffer): Result<Success>;
bind       (sockfd: FileDescriptor, addr: SocketAddr                  ): Result<Success>;
listen     (sockfd: FileDescriptor, backlog?: number                  ): Result<Success>;
shutdown   (sockfd: FileDescriptor, type: SocketShutOpt               ): Result<Success>;
getaddrinfo(node: string, service: string): Array<SocketAddr>;
//async calls
accept     (sockfd: FileDescriptor                  ): Promise<[remotefd: FileDescriptor, remoteaddr: SocketAddr]>;
connect    (sockfd: FileDescriptor, addr: SocketAddr): Promise<Result<Success>>;
recv       (sockfd: FileDescriptor,                   buffer: ArrayBuffer, length?: number): Promise<Result<number>>;
send       (sockfd: FileDescriptor,                   buffer: ArrayBuffer, length?: number): Promise<Result<number>>;
recvfrom   (sockfd: FileDescriptor,                   buffer: ArrayBuffer, length?: number): Promise<[total: Result<number>, from: SocketAddr]>;
sendto     (sockfd: FileDescriptor, addr: SocketAddr, buffer: ArrayBuffer, length?: number): Promise<Result<number>>;

```

functional tests done on Cosmo+Linux env:
- IPv4 TCP HTTP client
```js
const sockfd = must(os.socket(os.AF_INET, os.SOCK_STREAM));
await os.connect(sockfd, { addr: "51.15.168.198", port: 80}); // bellard.org
const httpReq = ["GET / HTTP/1.0", "", ""].join('\r\n')
await os.send(sockfd, Uint8Array.from(httpReq, c => c.charCodeAt(0)).buffer);
const chunk = new Uint8Array(512);
const recvd = await os.recv(sockfd, chunk.buffer);
console.log([...chunk.slice(0,recvd)].map(c => String.fromCharCode(c)).join(''));
```
- IPv6 UDP NTP client
```js
const sockfd = must(os.socket(os.AF_INET6, os.SOCK_DGRAM));
await os.sendto(sockfd, { addr: "2610:20:6f97:97::6", port: 37 }, new ArrayBuffer());
const u32 = new Uint32Array(1);
await os.recvfrom(sockfd, u32.buffer);
const seconds1900 = new DataView(u32.buffer).getUint32(0)
console.log(`${seconds1900} have passed since 1900 (${seconds1900/60/60/24/365} years)`);
```

What's bugging me:
- lot's of buf<->str needed (an `std.TextEncoder`/`std.TextDecoder` would be nice)
- quickjs shall provide it typing definition ([doc/os.d.ts](https://github.com/user-attachments/files/20739480/os.d.ts.txt) / `std.d.ts`) in it release archive
- calling 2 times a recv/send without await will create endless select() for the later
- @ceedriic you can play with the API (I can build a .exe/.elf if neede) I'm interested in parallel-handling HTTP server demo

yne:
@bellard done :heavy_check_mark: 

Also, returned `addrinfo[]` objects now also have an additional `socktype` property to be used to create sockets.

```js
const [addr] = must(os.getaddrinfo("bellard.org", { service: 80 }));
const sockfd = must(os.socket(addr.family, addr.socktype));
await os.connect(sockfd, addr);
```

yne:
@bellard shall I keep this PR alive ?

yne:
Following the `poll()` improvements in https://github.com/bellard/quickjs/commit/163b9b725dd60ad07b56877bcedaa994d6e3551e some socket-api are to be refreshed, @bellard do you feel up to the task ? (please say yes 🤞🏻 my imposter syndrome have never been so high since I've created this PR)

PR #416, opened 2025-06-05
Update Makefile

dobby33:
After enabling `CONFIG_WIN32` at line 35 in Makefile, command `make` still failing with such message:

_C:/w64devkit/bin/ld.exe: cannot find -ldl: No such file or directory_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,9 +243,10 @@ QJS_LIB_OBJS=$(OBJDIR)/quickjs.o $(OBJDIR)/dtoa.o $(OBJDIR)/libregexp.o $(OBJDIR
 
 QJS_OBJS=$(OBJDIR)/qjs.o $(OBJDIR)/repl.o $(QJS_LIB_OBJS)
 
-HOST_LIBS=-lm -ldl -lpthread
+HOST_LIBS=-lm -lpthread
 LIBS=-lm -lpthread
 ifndef CONFIG_WIN32
+HOST_LIBS+=-ldl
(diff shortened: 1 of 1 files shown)


nickva:
I had never tried w64devkit but on windows I compile it with  Msys2 like in the CI step:

https://github.com/bellard/quickjs/blob/638ec8ca5e1d4aed002a9fb3ef3358e2a6bc42ab/.github/workflows/ci.yml#L244-L250

Another alternative is to cross-compile on Linux for Windows:

https://github.com/bellard/quickjs/blob/638ec8ca5e1d4aed002a9fb3ef3358e2a6bc42ab/.github/workflows/ci.yml#L215-L229

dobby33:
`ld --version`
GNU ld (GNU Binutils) 2.44

nickva:
In the msys2 environment the `mingw-w64-ucrt-x86_64-dlfcn` provides the runtime dynamic library loading. It's a separate package that has to be installed.

dobby33:
> In the msys2 environment on windows the `mingw-w64-ucrt-x86_64-dlfcn` package provides the runtime dynamic library loading. It's a separate package that has to be installed.

I suspected that this was the case.

Although my assumption was incorrect. Using MSYS2 does apply `CONFIG_WIN32`. https://github.com/bellard/quickjs/blob/638ec8ca5e1d4aed002a9fb3ef3358e2a6bc42ab/Makefile#L88-L89

PR #433, opened 2025-08-21
Adjust `lastIndex` to leading surrogate when inside a surrogate pair in unicode RegExp

auvred:
22.2.7.2 RegExpBuiltinExec:

> 11. If fullUnicode is true, let input be StringToCodePoints(S). Otherwise, let input be a List whose elements are the code units that are the elements of S.

> 13.b. Let inputIndex be the index into input of the character that was obtained from element lastIndex of S.

See also:
- https://github.com/tc39/ecma262/issues/128
- https://chromium-review.googlesource.com/c/v8/v8/+/6368822


bellard:
applied

PR #476, opened 2026-01-01
illumos build fix

jwntree:


bellard:
I think it is better to maintain the patch separately as I cannot test with all the possible OSes.

PR #484, opened 2026-01-27
Support CC_FOR_BUILD for cross-compilation

chrisburr:
## Summary
Add support for cross-compilation environments that set `CC_FOR_BUILD` to specify the build machine compiler, complementing the existing `CROSS_PREFIX` approach.

## Motivation
When cross-compiling (e.g., building arm64 binaries on x86_64), the build needs to run `qjsc` to generate `repl.c`. The current `CROSS_PREFIX` mechanism works when the cross-compiler follows the `<prefix>gcc` naming convention, but many cross-compilation environments set `CC` directly and provide `CC_FOR_BUILD` for the native compiler.

## Changes
- Detect `CC_FOR_BUILD` and build `host-qjsc` using the native compiler
- Add `HOST_CFLAGS` and `HOST_LDFLAGS` to avoid passing target-specific flags (like `-mcpu=power8`) to the host compiler
- Skip building examples when cross-compiling (they require running `qjsc` to generate target executables)

## Testing
Tested cross-compiling for osx-arm64 and linux-ppc64le from x86_64 build machines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,6 +166,22 @@ CFLAGS_DEBUG=$(CFLAGS) -O0
 CFLAGS_SMALL=$(CFLAGS) -Os
 CFLAGS_OPT=$(CFLAGS) -O2
 CFLAGS_NOLTO:=$(CFLAGS_OPT)
+# Host flags for cross-compilation (avoid target-specific flags like -mcpu)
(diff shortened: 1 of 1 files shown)


dofaromg:
@copilot

PR #491, opened 2026-03-17
fix: make stack overflow detection work across thread/fiber context switches

davidroman0O:
`js_check_stack_overflow()` compares the current stack pointer against `stack_top`, which is captured once at runtime creation. Embedders using M:N threading (green threads, fibers, coroutines) call the runtime from execution contexts with independent stacks at unrelated addresses. The comparison produces false "stack overflow" errors because sp has no relationship to the original stack_top.

QuickJS already provides `JS_UpdateStackTop()` for this case (the header says "should be called when changing thread"), but in M:N schedulers the context switch is invisible to the embedder, there is no reliable hook to call it.

## Fix

Detect context switches inside `js_check_stack_overflow()` itself. If sp is above stack_top or far below stack_limit, the runtime is on a different stack. Re-anchor via `JS_UpdateStackTop()` before comparing.

- A genuine overflow: sp is just below stack_limit (grew past the limit)
- A context switch: sp is in a completely different address range

These two cases are distinguishable. The `unlikely()` branch is never taken on single-threaded use, so there is zero overhead for existing users.

## Testing

Verified with sequential runtime creation in Go (buke/quickjs-go):
- Before: second runtime fails with "SyntaxError: stack overflow"
- After: multiple sequential runtimes work correctly
- QuickJS test suite passes unchanged

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1638,6 +1638,17 @@ static inline BOOL js_check_stack_overflow(JSRuntime *rt, size_t alloca_size)
 {
     uintptr_t sp;
     sp = js_get_stack_pointer() - alloca_size;
+    /* Detect when the runtime is being called from a different thread or
+       fiber than the one that set stack_top. In M:N threading models
+       (green threads, fibers, coroutines), each execution context has its
+       own stack at an unrelated address. If sp is outside the expected
+       range, re-anchor stack_top to the current stack before checking.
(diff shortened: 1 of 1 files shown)


bellard:
I don't think it is a good idea to hide the logic inside the js_check_stack_overflow()

PR #506, opened 2026-04-22
Fix js_atod rounding when input has more than max_digits significant digits

### LINKED ISSUE #452: `100000000000000000000000.000000000000001` is `1e+23` instead of `1.0000000000000001e+23`
I bisected this to https://github.com/bellard/quickjs/commit/993660621a0a834c93eaad68713b2dc25508b326

Before this commit, quickjs matches other engines and prints `1.0000000000000001e+23`, but after, it prints `1e+23`.

(Found the same way as #449, but I mistakenly though it was a `ccgo` bug because I had not tested `master` here. Oops.)

jakebailey:
I know #452 was closed, but it seems like this is actually a pretty simple change when I went to look into it; just don't check `radix_bits` and set the bit regardless.

This would improve compatibility for these edge cases, which would make fuzz testing the number lib of TypeScript's Go port easier.

Fixes #452

### DIFF
--- a/dtoa.c
+++ b/dtoa.c
@@ -1505,8 +1505,10 @@ double js_atod(const char *str, const char **pnext, int radix, int flags,
     }
     
     /* Use the extra digits for rounding if the base is a power of
-       two. Otherwise they are just truncated. */
-    if (radix_bits != 0 && extra_digits != 0) {
+       two. Otherwise use them as a sticky bit so that round-half-to-even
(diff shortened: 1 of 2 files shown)


bellard:
It works for your example but not in the general case, so I don't know if it is worth doing the change.

jakebailey:
Interesting, do you have any tests in mind?

PR #510, opened 2026-04-30
Add SuppressedError class and resource management methods

ebrahimmostafa133:


bellard:
The coding style should be respected. Moreover, it is not really useful without more explicit resource management features.

PR #514, opened 2026-05-16
Fix leak in js_array_buffer_constructor3

bptato:
If alloc_flag is unset, then we are expected to consume the buffer even if we throw an exception.  (At least this is what js_array_buffer_transfer seems to assume.)


quickjs.c
```diff
@@ -56056,13 +56056,15 @@ static JSValue js_array_buffer_constructor3(JSContext *ctx,
     if (!alloc_flag && buf && max_len && free_func != js_array_buffer_free) {
         // not observable from JS land, only through C API misuse;
         // JS code cannot create externally managed buffers directly
```
bptato, line 56058: By the way, if this is a user error then maybe an assert would suffice?

bptato:
Actually, there's a better way to do this.  I'll open a new PR.

PR #516, opened 2026-05-27
Fix leak on allocation failure in ArrayBuffer#transfer

bptato:
Move construction /before/ the buffer is detached, so that an allocation failure doesn't detach the backing buffer.  (My previous attempt just cleaned up the buffer in this case; this patch preserves it so it can still be used after catching the InternalError.)

This time I've left JS_NewArrayBuffer unchanged, and instead documented its existing ownership semantics.  (I'm not sure if this is better, but at least there is no breaking change.)


bellard:
Fixed

